PR #40, opened 2018-02-26
fix serialization, internal issue #375

qoire:
Fixes:
* BlockInfo now uses RLP for default serialization and deserialization when possible
* BlockInfo deserializer can default to old implementation when necessary

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
@@ -25,12 +25,15 @@
 package org.aion.zero.impl.db;
 
 import org.aion.base.db.IByteArrayKeyValueDatabase;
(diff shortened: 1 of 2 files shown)


aion-jin:
BlockInfo constructor : 
     need check RLP.decode2(ser) if is emtpy before pull first element.

MigrationRedirectingInputStream:   
    throwing exceptions in the constructor to be bad practice

qoire:
Will patch the decoding logic to return early if the list is empty.

As for point 2) ``IOException`` is inherited from ``ObjectInputStream``.

aion-jin:
but keep in mind ,    the object serialization not only for blockinfo,  but also the outside wrapper,--> List<>,  which use ObjectDataSource<>,   if the wrapper was changed before we migrate to pub test net, we need review this patch.

PR #48, opened 2018-02-27
Issue365

AionJayT:
fix invalid nonce tx during the transactions coming to the txpool and doing many snapshots.


yulongnuco:
Needs to make the following changes to the `app.js`:

```
core.log("sending " + BATCH_THRESHOLD + " transfers ...");
await unlockAccount(owner);
```

PR #57, opened 2018-02-28
java api - batch data retrieval functions in admin api

ali-sharif:


modApiServer/src/org/aion/api/server/pb/ApiAion0.java
```diff
@@ -1306,6 +1224,185 @@ public void onBlock(IBlockSummary cbs) {
...
+                    count = 1000L;
+                }
+
+                Long latest = this.getBestBlock().getNumber();
+                List<Long> numbers = new ArrayList<>();
+                long i = Long.valueOf(latest);
+                while (i > Long.valueOf(latest)-Long.valueOf(count) && i >= 0) {
+                    numbers.add(i);
+                    i--;
+                }
```
AionJayT: you can replace with instead of while loop
LongStream.rangeClosed(start, end)
    .boxed().collect(Collectors.toList());

AionJayT: please refer
https://stackoverflow.com/questions/10242380/how-can-i-generate-a-list-or-array-of-sequential-integers-in-java/36114723#36114723

ali-sharif:
made changes requested; validated for regression by running testcase in javapi client.

ali-sharif:
made some more minor fixed. let me know of any more issues.

PR #58, opened 2018-02-28
initial block prop refactor, decouples from syncing #55

qoire:
This PR:

* Decouples block propagation entirely from SyncMgr
* Reworks some of the logic associated with block propagation

In particular, the rules for propagating a block are now stricter, in particular a block must be:

* Of higher or equal number to local
* Have a valid header
* Be a direct descendant of our local state

Send logic is also now more strict, a node will send to ALL peers that have a lower ``blockNumber`` than itself.


modAionImpl/src/org/aion/zero/impl/sync/BlockPropagationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        if (nodeId == null)
+            return Status.DROPPED;
+
+        if (block == null)
+            return Status.DROPPED;
+
+        ByteArrayWrapper hashWrapped = new ByteArrayWrapper(block.getHash());
+
+        // guarantees if multiple requests of same block appears, only one goes through
+        synchronized(this.cacheMap) {
```
aion-Ross: Possible to attack cache here?

Flood node with well formed but invalid blocks. Fill the cache with junk causing it to re-send valid but already seen blocks.

Flip order and validate header first before adding to the map?

qoire: Thanks,  I've moved the check for a well-formed header above the cache. This way at the very least and attack has to go through PoW to invalid the cache.

qoire:
Look at #61 for the same PR, but rebased

PR #61, opened 2018-02-28
rebase #55

qoire:
Same as #58, but rebased instead of merge


qoire:
From #58, as per @aion-Ross suggestion header validation is moved above caching, this way an attacker that wants to flood the cache must compute PoW making the attack cost inefficient.

PR #62, opened 2018-02-28
Fixing incorect link for the ID generation script

pfuhrmann:


pfuhrmann:
It was removed in dev

PR #68, opened 2018-03-01
Synchronize LRUMap

### LINKED ISSUE #65: LRU mapper in sync mgr is not thread safe, need add lock.
image

yulongaion:
There are two more raw LRUMap in `TransactionStore.java` and `BlockPropagationHandler.java`. The original authors have synchronized all access to the map, so there is no need to synchronize again.

Resolves #65


aion-jin:
Please put all LRU changes in one pull request.

Issue #74, opened 2018-03-01
Web3 getBlock, getWork, transactionLog issues


nuco-shidokht:
- boolean `fullTransactions = Boolean.parseBoolean((String) params.get(1));` throws an exception because it cannot convert types.
boolean` fullTransactions = Boolean.parseBoolean(params.get(1) + "");` works.

- When the boolean value is set to true `(eth.getBlock('blk_hash', true))`, blockToJson throws NullPointerException, because transaction is not 'contract creation' and contract address is set to Null in `jsonTx.put("address", TypeConverter.toJsonHex(_tx.getContractAddress().toString()));`
Also some fields required in Web3 formatting are missing.

- `eth_getWork` returns a byte array instead of the hash value.

- blockNumber, transactionIndex, logIndex fields are missing in transaction logs.

Corrections are in _api-dev_ branch.

ali-sharif:
resolution available in https://github.com/aionnetwork/aion/pull/75

ali-sharif:
Fixed by PR https://github.com/aionnetwork/aion/pull/75

PR #75, opened 2018-03-01
web3 server-side changes (JSON RPC spec compliance + bug-fixes)

ali-sharif:
courtesy of Shidokht.

fix for https://github.com/aionnetwork/aion/issues/74

### DIFF
--- a/modApiServer/src/org/aion/api/server/http/HttpServer.java
+++ b/modApiServer/src/org/aion/api/server/http/HttpServer.java
@@ -155,7 +155,7 @@ public final class HttpServer {
 
         case eth_getBlockByHash: {
             String hashString = (String) params.get(0);
-            boolean fullTransactions = Boolean.parseBoolean((String) params.get(1));
+                boolean fullTransactions = Boolean.parseBoolean(params.get(1) + "");
             if (hashString == null) {
                 log.debug("eth_getBlockByHash: invalid input");
                 return processResult(_id, null);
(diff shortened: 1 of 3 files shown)


modApiServer/src/org/aion/api/server/http/HttpServer.java
```diff
@@ -169,8 +169,7 @@ private static JSONObject process(final IRpc.Method _method, final long _id, fin
 
         case eth_getWork:
             // Header without nonce and solution , pool needs add new nonce
-            return processResult(_id, api.getBestBlock().getHeader().getHeaderBytes(true));
-
+                return processResult(_id, toHexString(HashUtil.h256(api.getBestBlock().getHeader().getHeaderBytes(true))));
```
ali-sharif: need review from @aion-6368726973 or @aion-Ross if this will break miner compatibility.

ali-sharif: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getwork

aion-Ross: At this time it should not break the reference miners/pool as they are relying on the "getblocktemplate" call.

However why are we returning a hash of the block header? If we want to update this to return Equihash work we would need eth_getWork to return the entire block header. Our result will be quite different from the referenced eth_getWork.

PR #97, opened 2018-03-02
P2pmgr policy based random node pick impl.

aion-jin:


aion-jin:
if we don't ask status of node which lower 128 blocks,  how we know those node 's status after they fully synced ?

PR #120, opened 2018-03-05
Persist nodes and metrics on disk

RobertSynapsics:


aion-jin:
Usually we need a testcase for new functions. Please add those junit class later on.

PR #129, opened 2018-03-06
Threads priority refactoring

serj-c:


aion-jin:
please provide me a top -H -p printscreen for aion thread list to prove it works well.
one majority part is set the thread pool's priority, which is not include in this patch.

serj-c:
OS related priority is the same, however the priority for sync-import is higher in thread scheduler in JVM. It is quite hard to see from the screenshots, but while testing / looking through console logs, there is a difference.

Results before my changes:
screenshot from 2018-03-07 15-21-31

Results after my changes:
screenshot from 2018-03-07 15-24-04

aion-jin:
stacktimer and p2p-in are not the place we can schedule to lower priority.  I'll close this PR till we have better plan.

PR #134, opened 2018-03-06
bugfix: on revert call, remove all blocks from old levels

AlexandraRoatis:


modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -390,8 +390,19 @@ public void revert(long previousLevel) {
         // walk back removing blocks greater than the given level value
         IAionBlock bestLine = bestBlock;
         while (currentLevel > previousLevel) {
+
+            // remove all the blocks at that level
+            List<BlockInfo> currentLevelBlocks = getBlockInfoForLevel(currentLevel);
+            if (currentLevelBlocks.size() == 0){
```
qoire: WIll this ever be true? given that we just grabbed the latest block from blockInfo

modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -390,8 +390,19 @@ public void revert(long previousLevel) {
...
+                for (BlockInfo bk_info : currentLevelBlocks) {
+                    blocks.delete(bk_info.getHash());
+                }
+            }
+
+            // remove the level
             index.remove((int) currentLevel);
-            blocks.delete(bestLine.getHash());
             bestLine = getBlockByHash(bestLine.getParentHash());
             --currentLevel;
```
yulongnuco: Should we start delete all levels from the max number to the revert number, level by level?

AlexandraRoatis: `getBestBlock()` uses this method to determine the height:
https://github.com/aionnetwork/aion/blob/dev/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java#L257-L261
so it will delete all levels higher than the given parameter

modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
AlexandraRoatis: since the BlockInfo for the block can be null, I think it is possible:
https://github.com/aionnetwork/aion/blob/35a24fbbeef1e74b9d7068ab12b8bf99f556d952/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java#L327-L332

but it's also possible that the `currentLevelBlocks` is null, so I'll add a null check

qoire: LGTM

PR #153, opened 2018-03-08
Index database keys from int to long

AlexandraRoatis:
* replaced `DataSourceArray.java` with `DataSourceLongArray.java` to avoid the type casting from long to int and back (which may cause problems in the future)
* added functionality for automatic conversion when using a database from an old kernel version
* minor cleanup: removed unused flushDangerously method and clone method


gonuco:
let's evaluate the impact before we merge.

PR #162, opened 2018-03-09
implemented a tx cache layer for tx propagation (issue125)

AionJayT:
1. Add a PendingTxCache class for cacheing the tx to make sure the tx going to pending pool by order.
2. remove unused StackTimer in PendingStateImpl.

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/blockchain/PendingTxCache.java
+++ b/modAionImpl/src/org/aion/zero/impl/blockchain/PendingTxCache.java
@@ -53,33 +53,34 @@ public class PendingTxCache {
         Map<BigInteger,AionTransaction> accountCacheMap = pendingTx.get(addr);
         if (accountCacheMap != null) {
             accountCacheMap.putAll(txmap);
-            rtn.add(accountCacheMap.get(bn));
 
-            boolean foundNext = true;
-            while(foundNext) {
-                bn = bn.add(BigInteger.ONE);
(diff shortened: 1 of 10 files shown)


modAionImpl/src/org/aion/zero/impl/blockchain/PendingTxCache.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class PendingTxCache {
+
+    private Map<Address, Map<BigInteger,AionTransaction>> pendingTx;
+    protected static final Logger LOG = AionLoggerFactory.getLogger(LogEnum.GEN.name());
+
+    public PendingTxCache() {
+        pendingTx = Collections.synchronizedMap(new LRUMap<>(1_000_000));
```
qoire: I feel like this value may be too big, heres my reasoning:

from https://ethereum.stackexchange.com/questions/1106/is-there-a-limit-for-transaction-size, at 3.1m gas the average block size is ~800kb, so if we apply the same logic to our limit, it means we can roughly be flooded with transactions that have a size of 2.4mb. This translates to a theoretical limit of 2.4 TB of memory use given an attacker incentivized to flood this cache.

Xiaoyang-Zhu:
I have a question, how you can confirm that the chain you are mining on is the aion main chain?

PR #190, opened 2018-03-15
Database clean-up

AlexandraRoatis:
* critical bug fixed: storage database was not closed on repository closing call
* fixed tests that were failing due to overlapping storage folders
* added compact() functionality to databases and repository
* applying compact after recovery with -r
* some code and messages clean-up

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/RecoveryUtils.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/RecoveryUtils.java
@@ -93,6 +93,7 @@ public class RecoveryUtils {
         store.pruneAndCorrect();
         store.flush();
 
+        ((AionRepositoryImpl) blockchain.getRepository()).compact();
         blockchain.getRepository().close();
     }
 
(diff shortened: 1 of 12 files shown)


modAionImpl/src/org/aion/zero/impl/db/RecoveryUtils.java
```diff
@@ -93,6 +94,9 @@ public static void pruneAndCorrect() {
         store.pruneAndCorrect();
         store.flush();
 
+        // compact database after the changes were applied
+        blockchain.getRepository().compact();
+
```
yulongnuco: Will the compaction take long after too many update?

modDbImpl/src/org/aion/db/impl/leveldb/LevelDB.java
```diff
@@ -161,6 +166,21 @@ public void close() {
...
 
+    @Override
+    public void compact() {
+        // acquire read lock
+        lock.writeLock().lock();
+
+        LOG.info("Compacting " + this.toString() + ".");
+
+        try {
+            db.compactRange(new byte[] { (byte) 0x00 }, new byte[] { (byte) 0xff });
```
yulongnuco: Minor: there are keys beyond `0xff`

modAionImpl/src/org/aion/zero/impl/db/RecoveryUtils.java
AlexandraRoatis: I tested it for 300k blocks and it was minor compared to the total pruning time.

PR #199, opened 2018-03-16
deadlock fix

JayT106:


yulongaion:
This PR solves deadlock issue among `BroadcastTxHandler`, `BlkHandler` and `API` threads. The race condition issue between API and pending state still exists.

PR #210, opened 2018-03-17
Bugfix for AionBlockchainImpl.getTransactionInfo(hash) + java-api: added filter to batch request

ali-sharif:
1. fix minor bug in java api + 
2. workaround for transactionStore.get(txHash) referencing blockhashes which have been removed from blocks DB


yulongaion:
LGTM.

@AlexandraRoatis do we need to correct the transaction database after reverting/pruning the blockstore?

PR #234, opened 2018-03-19
Functionality for saving runtime information to file

AlexandraRoatis:
1. Printing runtime reports for:
* p2p status
* the node's sync status
* block information for past K blocks (where K is given in the config file)

2. Configuration for enabling/disabling these runtime reports.

3. Correcting the patterns in .gitignore to not match unintended folders.

4. Functionality for enabling heap dumps.


yulongaion:
1. Can we use time, instead of block number, to trigger reports? The kernel may get stuck and won't be able to generate reports.
2. Heap dumps take too much time. Better replace it with thread dump.

yulongaion:
Also, since you're using JMX to query JVM info, do we need to add javax.management to the module list in launch script?

PR #240, opened 2018-03-20
add ExecutorService when the event method callback 

AionJayT:
Add ExecutorService for decreasing the multiple event callback time.


modEvtMgrImpl/src/org/aion/evtmgr/impl/handler/BlockHandler.java
```diff
@@ -55,19 +55,21 @@ public BlockHandler() {
...
+                es.submit(new Thread(() -> {
+                    switch (event.getCallbackType()) {
+                    case 0:
+                        ((EventCallbackA0) cb).onBlock(event.getFuncArgs().get(0));
+                        break;
+                    case 2:
+                        ((EventCallbackA0) cb).onBest(event.getFuncArgs().get(0), (List) event.getFuncArgs().get(1));
+                        break;
+                    default:
+                    }
```
yulongaion: Submit a `Thread` (even though it implements `Runable`) seems odd to me. Why not just passing the lambda `Runnable` to `ExecuteService`?

JayT106: it's just for the thread name. I can remove it.

PR #241, opened 2018-03-20
implement repay tx in pendingtximpl & pool 

AionJayT:
1. repay tx implement
2. refine log in txpool


yulongaion:
This may temporarily corrupt the pending state, but it will be corrected after a new block.

PR #251, opened 2018-03-21
Add block cache

yulongaion:


aion-jin:
smart change,  p2pmgr don't need query kernel for every incoming request,  for status,head and body,  only query kernel 1 times per second and send the cache value to peers.

PR #258, opened 2018-03-23
Evtmgr& eventcallback  system refactor

AionJayT:
1. simplified callback function implement
2. decoupled  evtmgr and callback registered classes
3. simplified eventHandlers, only have onEvent method
4. introduced EventExecuteService (EES) class for queuing events for the callback registered classes.


AionJayT:
merge error fixed

PR #261, opened 2018-03-23
Dev p2p write nobuffer

aion-jin:


aion-jin:
I don't have time to finish it with  lock . try lock   to replace the atomic variable with while , sleep.    could you enhance it when merge ?

aion-jin:
there are no lock free queue inside jdk,  you use msg queue just move blocking from one place to another place and use more memory.  msg queue is blocking queue, still block access when have concurrent access.

PR #285, opened 2018-03-26
Parallelize transaction validation; Optimize syncing

yulongaion:
* Parallelize transaction validation
* Stop requesting headers when the imported blocks queue is full
* Increase default database cache size
* Add block parent hash check


JayT106:
We can think to refine the Parallelize transaction validation part later.

Issue #284, opened 2018-03-26
DB Configurations: block_size, max_fd_alloc_size


qoire:
Adds in two configuration parameters to database, these are applied on every database:

* ``block_size`` - configures the block size of the underlying database, this parameter is levelDB specific. This parameter should be specified as: <number><sizePrefix>, where sizePrefix can be ``B``, ``mB``, ``MB``, ``M``, ``G``, ``gB``, ``GB``, ``K``, ``kB``, ``KB``, depending on your desired file size. Please note that LevelDB implicitly caps this filesize between 1024B (1kB) to 4mB.

* ``max_fd_alloc_size`` refers to the ``maxOpenSize`` parameter on levelDB options, this is applied per database, so setting a value of 1024 requires a global file descriptor minimum of n * (1024) where n is the number of databases opened. It is up to the node sysadmin to configure this on the system properly.

qoire:
As requested by @yulongaion, additional parameters will be added:

* ``cache_size``
* ``write_buffer_size``

Both parameters will use the same parsing format as specified in ``block_size``

PR #286, opened 2018-03-26
DB Config changes for #284

qoire:
This PR will apply the changes noted in #284 , we still need to benchmark the DB. Please do not close the issue yet as there are still some refinements that must be placed, namely:

* Proper handling of parse errors (currently they are silently ignored)
* Proper tuning of LevelDB

**Subtleties to note**, with this change, old config files will be missing two entires:

* ``max_fd_alloc_size``
* ``block_size``

These should be set in the ``db`` portion of the config. Refer to the default config file in ``modBoot/resource/config.xml``

In the absence of these two parameters, ``CfgDb`` will attempt to use default values:
* max_fd_alloc_size = 1024
* block_size = 16kB (16 * 1024)

These two values may not be the optimal, I have not tuned them yet! So please set these parameters when you update. Recall that the original parameters were:

* max_fd_alloc_size = 32
* block_size = 10mB (10 * 1024 * 1024)

### DIFF
--- a/modAionBase/src/org/aion/base/db/IRepositoryConfig.java
+++ b/modAionBase/src/org/aion/base/db/IRepositoryConfig.java
@@ -71,4 +71,8 @@ public interface IRepositoryConfig {
      * @return
      */
     String getMaxHeapCacheSize();
+
+    int getMaxFdAllocSize();
+
+    int getBlockSize();
 }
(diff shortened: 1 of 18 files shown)


qoire:
Closing, due to failing CI tests. Will resubmit when updates to #284 and CI is fixed

PR #306, opened 2018-03-28
pendingState buffer implement

AionJayT:
1. Add buffer layer when the tx finished the VM pre-check
2. patch null check in the txpool remove method
3. remove unnecessary synchronized
4. let bestblock be an atomic class.
5. move pendingState LOG category from GEN to TX 
6. txpool fee pick algo revise.
7. fix pendingCacheTx size calculation bug


modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
```diff
@@ -398,15 +471,20 @@ private boolean addPendingTransactionImpl(final AionTransaction tx, BigInteger t
...
             }
 
-            AionTransaction rtn = this.txPool.add(tx);
-            if (rtn != null && !rtn.equals(tx)) {
-                AionTxReceipt rp = new AionTxReceipt();
-                rp.setTransaction(rtn);
-                receivedTxs.remove(ByteArrayWrapper.wrap(rtn.getHash()));
-                fireTxUpdate(rp, PendingTransactionState.DROPPED, getBestBlock());
+            if (bufferEnable) {
+                txBuffer.add(txSum);
```
qoire: This may get expensive if we update on every transaction

modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
```diff
@@ -226,14 +292,12 @@ private int getPendingTxSize() {
...
-        return this.txPool.snapshot();
+            return this.txPool.snapshot();
     }
 
     public synchronized AionBlock getBestBlock() {
-        if (best == null) {
-            best = blockchain.getBestBlock();
-        }
-        return best;
+        best.set(blockchain.getBestBlock());
```
qoire: Are there any implications of changing the logic from a lazy load to something that is grabbed every time?

JayT106: For now, only when Api create the block template will call this method. Inside the pendingState doesnt use this method anymore. the previous implement in the pendingState call this method anywhere.

replace to the Collections.synchronizedList (new ArrayList<>() )

modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
JayT106: you mean the CopyOnWriteArrayList? It can be refactored later.

AionJayT: replace to the Collections.synchronizedList (new ArrayList<>() )

PR #308, opened 2018-03-28
Cleaning up synchronization in the repository tracker

AlexandraRoatis:
* added separate locks for changes to accounts and changed to contracts
* removed synchronization from all methods

### DIFF
--- a/modMcf/src/org/aion/mcf/db/AbstractRepositoryCache.java
+++ b/modMcf/src/org/aion/mcf/db/AbstractRepositoryCache.java
@@ -18,17 +18,16 @@
  * Contributors:
  *     Aion foundation.
  *******************************************************************************/
-
 package org.aion.mcf.db;
 
 import org.aion.base.db.IContractDetails;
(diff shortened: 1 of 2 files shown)


modMcf/src/org/aion/mcf/db/AbstractRepositoryCache.java
```diff
@@ -60,130 +61,203 @@
...
 
     @Override
-    public synchronized void deleteAccount(Address address) {
-        getAccountState(address).delete();
-        getContractDetails(address).setDeleted(true);
+    public void deleteAccount(Address address) {
+        lockAccounts.writeLock().lock();
+        lockDetails.writeLock().lock();
+        try {
+            getAccountState(address).delete();
```
qoire: My impression is that this wouldn't work, due to getAccountState() requiring a ``readLock``.

qoire: Nevermind I am incorrect on this point.

PR #321, opened 2018-04-01
issue 319 fix

AionJayT:
1. eventmgr handler refine and onEvent method add try catch
2. add more try catch in p2pMgr task-inbound, send method


qoire:
Hey, two general notes: Can you expand your catch to ``Throwables`` for all catch clauses? Just catching exceptions wont protect us from RuntimeExceptions. Secondly (since you're working on this) is it possible to give us a catalogue of all long-run/critical/cannot-throw-exception threads? This will help us debug the kernel in the future.

PR #323, opened 2018-04-02
P2p Refactor

qoire:
This change will:

* Move IO onto a single thread
* Disable task reporting (as a merge artifact will fix at a later date)
* Enable a system where OP_WRITEs are used to trigger writes in the queue instead of looping
* Enable some filtering logic for ResStatusHandler (this is not essential we can remove this if necessary)

There are some artifacts, for example exposing peer status as an observer. They are harmless and disabled for now, if necessary I will remove them. You can safely ignore anything outside of ReqStatusHandler, P2pMgr and the new classes added for P2p.

ALSO: Im aware that ``nanoTime()`` calls are quite expensive, if necessary we can remove them.

Please read the logic carefully and correct any places where I made have made mistakes. It is imperative that multiple people are aware of the current thread model in place. I will not be re-enabling the file output for reports, if you wish to re-enable it please submit a future PR.


modP2pImpl/src/org/aion/p2p/impl/selector/MainIOLoop.java
```diff
@@ -0,0 +1,317 @@
...
+                    break;
+
+            if (this.needsToSelectAgain) {
+                selectAgain();
+                Set<SelectionKey> keys = this.currSelector.selectedKeys();
+                if (keys.isEmpty())
+                    break;
+                else
+                    it = selectedKeys.iterator();
+            }
```
yulongaion: Should this if statement at the beginning of the while loop?

modAionImpl/src/org/aion/net/Peer.java
```diff
@@ -0,0 +1,98 @@
...
+            rwTxLock.readLock().unlock();
+        }
+    }
+
+
+    // TODO: implementation of serialization, so we can store this
+    public byte[] getEncoded() {
+        return null;
+    }
+}
```
yulongaion: Are we switching from `one cache for all nodes` to `one cache for each node`?

modAionImpl/src/org/aion/zero/impl/sync/TaskGetStatus.java
```diff
@@ -76,8 +76,7 @@ public void run() {
             try {
                 for (int id : ids) {
                     p2p.send(id, reqStatus);
-
-                    Thread.sleep(Math.max(intervalMin, intervalTotal / ids.size()));
+                    Thread.sleep(1000L);
```
yulongaion: This could potentially affect the syncing speed and node last message timestamp.

modP2pImpl/src/org/aion/p2p/impl/NodeMgr.java
```diff
@@ -311,13 +323,16 @@ void moveInboundToActive(int _channelHashCode, final P2pMgr _p2pMgr) {
         if (node != null) {
             node.setConnection("inbound");
             node.setFromBootList(seedIps.contains(node.getIpStr()));
-            INode previous = activeNodes.putIfAbsent(node.getIdHash(), node);
+            INode previous = activeNodes.put(node.getIdHash(), node);
             if (previous != null)
-                _p2pMgr.closeSocket(node.getChannel());
+                _p2pMgr.closeSocket(((Node) previous).getChannel());
```
yulongaion: Same as above

modP2pImpl/src/org/aion/p2p/impl/P2pMgr.java
```diff
@@ -717,23 +504,27 @@ public void run() {
...
-
-                private AtomicInteger cnt = new AtomicInteger();
-
-                @Override
-                public Thread newThread(Runnable r) {
-                    return new Thread(r,"p2p-worker-" + cnt.incrementAndGet());
-                }
-            });
+            workers = Executors.newFixedThreadPool(Math.min(Runtime.getRuntime().availableProcessors() * 2, 16));
+            // workers = Executors.newCachedThreadPool();
```
yulongaion: Why we stop naming IO threads?

modP2pImpl/src/org/aion/p2p/impl/P2pMgr.java
```diff
@@ -852,16 +625,294 @@ public void shutdown() {
...
+            } finally {
+                if (byteBuffers.isEmpty())
+                    key.interestOps(SelectionKey.OP_READ);
+            }
+        }
+
+        @Override
+        public void acceptMessage(SelectableChannel channel, SelectionKey key, ByteBuffer buffer) {
+            byteBuffers.add(buffer);
+            key.interestOps(SelectionKey.OP_WRITE);
```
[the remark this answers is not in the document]
qoire: Registering ``OP_WRITE`` at initialization causes the selector to spin (99% CPU usage) since the channel is almost always ready for writing see:

http://rox-xmlrpc.sourceforge.net/niotut/

modAionImpl/src/org/aion/net/Peer.java
qoire: Id like to do that in the future, currently this design can be scrapped if we don't like it. It makes sense (in my mind) to manage state on a per node basis

modAionImpl/src/org/aion/zero/impl/sync/TaskGetStatus.java
qoire: I will be looking into refactoring sync in a separate target, this can be reverted if necessary

modP2pImpl/src/org/aion/p2p/impl/P2pMgr.java
qoire: Oops, this was a mistake on my part please add it back

modP2pImpl/src/org/aion/p2p/impl/P2pMgr.java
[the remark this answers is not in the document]
yulongaion: :+1:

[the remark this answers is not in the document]
yulongaion: But we're not updating the `interestOps` after finishing writing, right?

yulongaion:
Yao's too busy to apply the changes. I will merge and update.

PR #338, opened 2018-04-03
splitting the compact from the -r functionality

AlexandraRoatis:
The `./aion.sh -r` functionality walks through the main chain from the topmost block through links to parents till genesis. It removes the side chains and recomputes the total difficulty for each block. 

Because it deletes data from the database, a compact call is applied after the cleanup described above.

This PR separates the compact functionality from the block info cleanup. Now each can be called separately.


qoire:
Would be good to get some information about the split in the PR description

PR #353, opened 2018-04-05
heuristic for skipping tryToConnect for obviously large blocks

AlexandraRoatis:
The p2p workers need not acquire locks on the blockchain to do `tryToConnect` when the block is obviously much larger and will return `NO_PARENT`.


yulongaion:
Can we also check both the upper and lower bounds? Suggest `[-32,32]` for block propagation.

PR #366, opened 2018-04-06
BroadcastTxHandler buffer implement

AionJayT:
1. use timer to control the tx import.
2. p2p config settings add "txbroadcastbuffer" boolean option

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
+++ b/modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
@@ -66,15 +66,23 @@ public final class BroadcastTxHandler extends Handler {
 
     private LinkedBlockingQueue<AionTransaction> txQueue;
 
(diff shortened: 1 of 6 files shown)


modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
```diff
@@ -77,22 +114,31 @@ public final void receive(int _nodeIdHashcode, String _displayId, final byte[] _
...
+                log.trace("<BroadcastTxHandler from: {} empty {}>", _displayId);
 
             }
             return;
         }
-        pendingState.addPendingTransactions(castRawTx(broadCastTx));
+
+        if (this.buffer) {
+            try {
+                txQueue.addAll(castRawTx(broadCastTx));
```
yulongaion: Suggest changing to non-blocking add

AionJayT: add and offer is same inside blocking method implemented by the blocking queue

    public boolean add(E e) {
        if (offer(e))
            return true;
        else
            throw new IllegalStateException("Queue full");
    }

PR #376, opened 2018-04-11
Javaapi refactor

AionJayT:
1. remove unnecessary variable assign
2. modify zmq thread#
3. encapsulation class method


modApiServer/src/org/aion/api/server/ApiAion.java
```diff
@@ -649,20 +646,20 @@ public boolean setReportedHashrate(String hashrate, String clientId) {
...
     // leak the oracle instance. NrgOracle is threadsafe, so safe to do this, but bad design
-    public NrgOracle getNrgOracle() {
+    protected NrgOracle getNrgOracle() {
         return this.nrgOracle;
     }
 
-    public long getDefaultNrgLimit() {
-        return DEFAULT_NRG_LIMIT;
+    protected long getDefaultNrgLimit() {
+        return 500_000L;
```
ali-sharif: @qoire - can you double check if this number is good enough for most transactions?

AionJayT: 500_000 is equalto the DEFAULT_NRG_LIMIT, cause no other place use it. so I just remove the define. But I don't get what's the API for?

ali-sharif: This was in the api to comply with json-rpc spec which allows people to send transaction objects to the kernel without the gasLimit defined, so we have to pick some maximum limit on gas spending on their behalf.

PR #386, opened 2018-04-12
adding a drop database method

AlexandraRoatis:


yulongnuco:
Can we just destroy the database instead of deleting keys one-by-one?

PR #393, opened 2018-04-13
Alt sync

qoire:
Sync algo could potentially be abused, perhaps we could calculate the largest subset of common peers that are within a certain block distance from eachother.


qoire:
Sync strategy too risky, merged the other portion (block prop) in through cherry-pick

PR #341, opened 2018-04-04
Adding State to Sync

qoire:
As requested, this PR will:

* Track each peer based on their current state, this prevents us from repeatedly requesting headers and bodies, letting us "play nice" with peers.

PR #395, opened 2018-04-13
Introduce syncing modes

yulongaion:
This PR introduces three syncing modes, to help the network recover from side-chains:

- `NORMAL`: the peer is on main chain; sync with normal strategy
- `BACKWARD`: the peer is on side chain; try to search for the fork point
- `FORWARD`: the peer is on side chain; we've already figured out the fork point; sync from the fork point to the head.

**Pros:** 
- Allow nodes to re-branch at any height based on total difficulty

**Cons:**
- Increase storage/state size
- Potential double-spending

**TODO:** 
- Optimize the way we search for the fork point
- Integrate peer syncing state to avoid redundant requests, see #341

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/sync/TaskGetStatus.java
+++ b/modAionImpl/src/org/aion/zero/impl/sync/TaskGetStatus.java
@@ -44,11 +44,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 final class TaskGetStatus implements Runnable {
 
 
-    private final static int intervalTotal = 1000;
+    private final static int interval = 2000; // two seconds
 
-    private final static int intervalMin = 100;
-
-    // single instance req status
     private final static ReqStatus reqStatus = new ReqStatus();
 
     private final AtomicBoolean run;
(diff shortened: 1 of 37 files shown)


modAionImpl/src/org/aion/zero/impl/sync/TaskGetStatus.java
```diff
@@ -71,19 +68,24 @@
...
         while (this.run.get()) {
-            Set<Integer> ids = new HashSet<>(p2p.getActiveNodes().keySet());
-
             try {
+                Set<Integer> ids = new HashSet<>(p2p.getActiveNodes().keySet());
+
                 for (int id : ids) {
                     p2p.send(id, reqStatus);
-                    Thread.sleep(1000L);
+                    Thread.sleep(interval / ids.size());
```
qoire: Quick question, on the weekend we discussed a change of TaskGetStatus to:

~~~~java
for (int id : ids) {
  p2p.send(id, reqStatus);
}
Thread.sleep(<someTimeout>)
~~~~

Meaning status requests are sent out as a batch. Would this impact the algorithm in any discernable way?

modApiServer/src/org/aion/api/server/pb/ApiAion0.java
```diff
@@ -908,7 +908,7 @@ public ApiAion0(IAionChain ac) {
             try {
                 Message.rsp_syncInfo rsp = Message.rsp_syncInfo.newBuilder().setChainBestBlock(sync.chainBestBlkNumber)
                         .setNetworkBestBlock(sync.networkBestBlkNumber).setSyncing(!sync.done)
-                        .setMaxImportBlocks(sync.blocksRequestMax).build();
+                        .setMaxImportBlocks(0).build();
```
qoire: Earlier I saw that we requested in batches of 24, could we make that a constant and set it here as well?

PR #399, opened 2018-04-16
APIs for production use

qoire:
This PR will:

* Add some non-standard APIs under the private category to make my life easier

Equivalent client side:
https://github.com/qoire/aion_web3/tree/priv

Note: these APIs are non standard and should be considered unsupported for all intents and purposes. They may change at any time and are not guaranteed to work. Do NOT depend on them.

### DIFF
--- a/modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
@@ -1207,7 +1207,7 @@ public class ApiWeb3Aion extends ApiAion {
             return new RpcMsg(JSONObject.NULL);
 
         JSONObject tx = Tx.InfoToJSON(transaction,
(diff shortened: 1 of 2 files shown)


modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
```diff
@@ -1123,6 +1127,151 @@ else if (_params instanceof JSONObject) {
...
+        // begin output processing
+        AionTxInfo transaction = this.ac.getAionHub()
+                .getBlockchain()
+                .getTransactionInfo(hash);
+
+        if (transaction == null)
+            return new RpcMsg(JSONObject.NULL);
+
+        JSONObject tx = Tx.InfoToJSON(transaction,
+                this.ac.getBlockchain().getBlockByHash(transaction.getParentBlockHash()));
```
ali-sharif: this should be transaction.getBlockHash()

ali-sharif:
Looks good to me.

PR #409, opened 2018-04-17
PendingStateImpl tune-up and tx broadcast rate control adjustment

AionJayT:
1. decrease the tx flooding number from cached layer to pool layer. 
2. adjust tx broadcast rate control in p2p module 10/s -> 20/s base on the max buffer send in the pendingStateImpl.
3. refine log

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
+++ b/modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
@@ -158,11 +158,9 @@ public final class BroadcastTxHandler extends Handler {
         for (byte[] raw : broadCastTx) {
             try {
                 AionTransaction tx = new AionTransaction(raw);
-                System.out.println("castRawTx nonce: " + tx.getNonceBI().toString());
                 if (tx.getHash() != null) {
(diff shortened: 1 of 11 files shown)


modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
```diff
@@ -167,9 +167,14 @@ public final void receive(int _nodeIdHashcode, String _displayId, final byte[] _
                 }
             } catch (Exception e) {
                 // do nothing, invalid transaction from bad peer
+                System.out.println("castRawTx exception: " + e.toString());
```
qoire: Suggest implementing this in a logger, and set the logging level to debug or trace. Otherwise I think people may complain

qoire:
LGTM

modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
AionJayT: 2ad727b9227bf7cea433d2ee65b72c6930aa5dc1

PR #414, opened 2018-04-17
Api Priv Updates

qoire:
This PR will:

* Implement function for quickly showing important stats
* Implement function for showing peer sync status
* Further improve on function for showing peer p2p status
* Implement function for showing node configuration

Edit: Added unrelated change

* Changes the default build behaviour of pack_build, specifically ``prepack.sh`` will not grab the java runtime from the local ``$JAVA_HOME`` defined JDK, not the downloaded version


ali-sharif:
lgtm.

modAionImpl/src/org/aion/zero/impl/sync/SyncMgr.java
```diff
@@ -312,4 +312,13 @@ public synchronized void shutdown() {
         start.set(false);
         workers.shutdown();
     }
+
+    /**
+     * Reasoning behind only returning iterator is that we
+     * don't want other parties to unsafely modify the mapping
+     * in an unsafe fashion
+     */
+    public Iterator<Map.Entry<Integer, PeerState>> getPeerStatesIterator() {
```
ali-sharif: Instead of leaking the iterator (safe, since won't yield concurrent modification exceptions, but behaviour undefined (see https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html)). 

Maybe better to just return a copy using copy constructors:
```
Set<K> keySetCopy = new HashSet<>(map.keySet());
List<V> valuesCopy = new ArrayList<>(map.values());
```

qoire:
@ali-sharif updated as per the request

PR #421, opened 2018-04-18
make filters-enabled as hidden config

ali-sharif:
people started playing with it, causing web3 client not working correctly (as expected)


qoire:
Can we also remove it from the default config generation as well:

https://github.com/aionnetwork/aion/blob/6bbff7286519593a3f18d97f084dcbe4c0bdd59d/modMcf/src/org/aion/mcf/config/CfgApiRpc.java#L163-L168

qoire:
My mistake, was already included in the PR

PR #436, opened 2018-04-20
Heap caching for reads

AlexandraRoatis:
Preliminary tests show that 99.99% of reads will be done from the light cache for **block** and **index** (tested with a sync from block 0 to 10000). 

For **details** 99.55% of reads should be from LRU cache (tested with a sync from block 0 to 24000).


qoire:
@AlexandraRoatis hey, is it possible to rebase this onto the ``platypus`` branch, and possibly run a node with this setup configured? I'd like to set some unchangeable default values instead of adding more configuration entries.

gonuco:
Let's leave this PR  as Q2 patch after mainnet. Need more time testing.

PR #443, opened 2018-04-23
add write attempts checker

aion-6368726973:
add write attempts checker in order to make decision if current socket healthy to write in logic to prevent write threads hanged by channel


gonuco:
how you know write attempt > 2 is a broken pipe ?

PR #445, opened 2018-04-23
Rename syncSeedOnly to clusterNodeMode

iamyulong:
New configurations:
```xml
<p2p>
  <!-- If enabled, this node only actively connects to the seed nodes but accepts any inbound connections -->
  <cluster-node-mode>false</cluster-node-mode>
  <!-- If enabled, this node won't respond to block propagation, block headers/bodies requests -->
  <sync-only-mode>false</cluster-only-mode>
</p2p>
```

This is originally from @ali-sharif.


qoire:
@ali-sharif can I get you to confirm as well

Issue #441, opened 2018-04-23
WARN  API  [BlkHdr]: ExecutorService Q is full!


ghost:
On one of my computers after it had downloaded 57GB worth of the testnet blockchain and the kernel was still busy syncing (but must have been stuck, because the kernel on my other computer on a slower connection got to 65G overnight and they both started at around 53G), it kept showing this message

```
18-04-23 05:14:53.628 INFO  CONS [miner_status]: Aion internal miner generating 0.0 solutions per second
18-04-23 05:14:59.954 WARN  API  [BlkHdr]: ExecutorService Q is full!
18-04-23 05:14:59.954 INFO  SYNC [sync-ib]: <import-status: node = c33d49, hash = 88ffb4, number = 83059, txs = 106, result = IMPORTED_BEST, time elapsed = 8866 ms>
18-04-23 05:15:08.629 INFO  CONS [miner_status]: Aion internal miner generating 0.0 solutions per second
18-04-23 05:15:09.342 INFO  SYNC [sync-ib]: <import-status: node = c33d5f, hash = 9d012b, number = 83060, txs = 96, result = IMPORTED_BEST, time elapsed = 9386 ms>
18-04-23 05:15:09.342 WARN  API  [BlkHdr]: ExecutorService Q is full!
18-04-23 05:15:09.756 INFO  SYNC [p2p-worker-2]: <import-status: node = 6bd088, hash = 82c5ab, number = 95585, txs = 85, result = NOT_IN_RANGE>
18-04-23 05:15:14.123 INFO  SYNC [sync-ib]: <import-status: node = c33d39, hash = 55859c, number = 83061, txs = 59, result = IMPORTED_BEST, time elapsed = 4776 ms>
18-04-23 05:15:14.123 WARN  API  [BlkHdr]: ExecutorService Q is full!
18-04-23 05:15:18.339 INFO  SYNC [sync-ib]: <import-status: node = c33d39, hash = 285b50, number = 83062, txs = 51, result = IMPORTED_BEST, time elapsed = 4216 ms>
18-04-23 05:15:18.339 WARN  API  [BlkHdr]: ExecutorService Q is full!
18-04-23 05:15:19.118 INFO  SYNC [sync-ib]: <import-status: node = c33d39, hash = 98d9f0, number = 83063, txs = 11, result = IMPORTED_BEST, time elapsed = 779 ms>
18-04-23 05:15:19.118 WARN  API  [BlkHdr]: ExecutorService Q is full!
[...]

PR #448, opened 2018-04-23
Added low storage space suggestion when ExecutorService Q is full.

lope:
I resolved my issue.
https://github.com/aionnetwork/aion/issues/441

Here's a pull request if you like my suggestion to improve the error msg.

I discovered that when you're running out of storage you can experience this error. It's hard for a user to know what's wrong with the current error message, so this friendly question should help people who experience this issue, to resolve it quickly.


AionJayT:
Hi Lope, the executor service queue full doesn't have the relationship with the low storage space.
So we can't approve this PR.

PR #454, opened 2018-04-27
patch task in

aion-6368726973:
exception catch

### DIFF
--- a/modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
+++ b/modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
@@ -168,104 +168,108 @@ public final class P2pMgr implements IP2pMgr {
                     final SelectionKey sk = keys.next();
                     keys.remove();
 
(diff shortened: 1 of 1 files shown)


yulongaion:
Can we catch all exceptions for potential risks?

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -168,104 +168,108 @@ public void run() {
...
-
-                        } catch (CancelledKeyException e) {
-                            chanBuf.isClosed.set(true);
-                            closeSocket((SocketChannel) sk.channel(),
-                                    chanBuf.displayId + "-read-msg-key-cancelled-exception");
-                        } catch (Exception e) {
-                            if (showLog)
-                                System.out.println("<p2p-pi global exception>");
                         }
+                    } catch(Exception ex) {
```
qoire: We should log something here

PR #456, opened 2018-04-27
Database recovery

AlexandraRoatis:
check and recover db data when importing existing block if the block is missing the state or index

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
@@ -1411,9 +1411,6 @@ public class AionBlockchainImpl implements IAionBlockchain {
             }
         } while (!repo.isIndexed(other.getHash(), other.getNumber()));
 
-        // sync to the last correct state
-        repo.syncToRoot(other.getStateRoot());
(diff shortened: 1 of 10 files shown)


AlexandraRoatis:
The broken database is in `/Aion/AionFiles/database-invalid-block.tar.gz`. Working on additional unit tests.

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -1304,60 +1311,142 @@ public IEventMgr getEventMgr() {
...
+
+        long blockNumber = block.getNumber();
+        LOG.info("Corrupt world state at block hash: {}, number: {}."
+                         + " Looking for ancestor block with valid world state ...", block.getShortHash(), blockNumber);
+
         AionRepositoryImpl repo = (AionRepositoryImpl) repository;
 
-        Map<Long, AionBlock> dirtyBlocks = new HashMap<>();
-        AionBlock other;
+        Stack<AionBlock> dirtyBlocks = new Stack<>();
```
qoire: Prefer ``deque`` instead. Not critical

PR #457, opened 2018-04-27
fix issue#455

AionJayT:
When the seed node enabled, the mining relative functions should disable.


qoire:
Assigning @AlexandraRoatis, as Yulong may be too busy to review

PR #462, opened 2018-05-03
fix zmq workThread got stuck been occupied by the long data pull API

AionJayT:
issue:
Java API worker thread will get stuck when the client doing the heavy data query from the kernel, Therefore, the heartbeat message cannot get into the kernel and response.

solution:
separate the channel for heartbeat msg and the regular api message.

additional:
some small refactoring in the code.


ali-sharif:
will test and review before tommorow morning.

PR #464, opened 2018-05-04
update to blockdetails api endpoint for effecient range query 

ali-sharif:
+ added error to txDetail


modApiServer/src/org/aion/api/server/pb/ApiAion0.java
```diff
@@ -1483,6 +1483,149 @@ public ApiAion0(IAionChain ac) {
...
+                if (_blkStart < 0)
+                    blkStart = 0;
+                else
+                    blkStart = _blkStart;
+
+                // blocks requested in the future. return empty result
+                if (blkStart > latestBlkNum) {
+                    Message.rsp_getBlockDetailsByRange rsp =
+                            Message.rsp_getBlockDetailsByRange.newBuilder()
+                                    .addAllBlkDetails(new ArrayList<>())
```
qoire: Suggestion, could use static var here

Issue #416, opened 2018-04-18
aion.sh not working with symlinks


centrys-otniel:
The aion core can't be started from a symbolic link.

Suggestion: 
on line 3 in aion.sh, replace
`cd "$(dirname $0)"`
with
`cd "$(dirname $(realpath $0))"`

PR #468, opened 2018-05-08
Fix issue #416 and google-java jar added

aionick:


qoire:
Is there any internal reference a user can use to setup the linting/formatting process?

aionick:
The pre-commit file should be moved into the user's local .git/hook directory. The .git directory is local only so I've stored this file in .gitsettings for now

yulongaion:
Can we integrate the formatting thing into ant, e.g. `ant format`, so we can check code style any time?

AionJayT:
@yulongaion good idea, how about creating a ticket for add ant format task, then we do another PR?

Issue #298, opened 2018-03-28
Reenable JournalPruningDataSource after thorough testing


AlexandraRoatis:
JournalPruningDataSource was removed in #293 and #297.

PR #478, opened 2018-05-11
Enabling JournalPruneDataSource

AlexandraRoatis:
## Description

- minor cleanup in `BlockchainDataRecoveryTest.java`
- re-enabling the use of `JournalPruneDataSource.java` for Trie pruning
- configuration setting for the number of recent blocks to be maintained in the database (with a minimum value set to 128)
- utils for exploring the state trie
- functionality for rebuilding the genesis block when recovery is triggered by a missing state at kernel startup
- unit tests for the `JournalPruneDataSource.java` data storage and retrieval operations

Fixes Issue #298.

## Type of change

- [ ] Bug fix.
- [x] New feature.

## Testing

In addition to the unit tests above:
- the kernel was run with different settings of pruneBlockCount (10, 100, 1000, 10000) for over 24 hours, only the version keeping only 10 blocks resulted in issues, which are now avoided due to the minimum value setting
- ran the kernel with negative and zero pruneBlockCount and can confirm pruning disabled
- unit tests show that values are not cached when the pruning is disabled (note: the class was initially disabled due to memory usage concerns)
- ran the kernel after reverting to blocks without stored state and confirmed that the genesis was rebuilt and full state recovery was successful

## Verification


modMcf/src/org/aion/mcf/trie/JournalPruneDataSource.java
```diff
@@ -152,31 +158,35 @@ private Ref decRef(ByteArrayWrapper keyW) {
...
                 Ref ref = refCount.get(key);
                 if (ref == null || ref.journalRefs == 0) {
-                    batchRemove.put(key.getData(), null);
+                    batchRemove.add(key.getData());
                 } else if (ref != null) {
                     ref.dbRef = false;
                 }
             }
-            src.putBatch(batchRemove);
+            src.deleteBatch(batchRemove);
```
AionJayT: how many keys need to be removed during the prune? if it's a lot, use parallel-stream to shorten the processing time instead of the for loop?

modMcf/src/org/aion/mcf/trie/JournalPruneDataSource.java
```diff
@@ -244,12 +259,21 @@ public void commitBatch() {
 
     @Override
     public void deleteBatch(Collection<byte[]> keys) {
-        throw new UnsupportedOperationException();
+        if (!enabled) {
+            return;
+        }
+        // deletes are delayed
+        keys.forEach(key -> currentUpdates.deletedKeys.add(new ByteArrayWrapper(key)));
```
AionJayT: same here. can we do parallel?

AlexandraRoatis:
I think the batch sizes are too small to need parallelism here at present.

modAionImpl/src/org/aion/zero/impl/AionHub.java
```diff
@@ -272,9 +270,23 @@ private void loadBlockchain() {
...
             byte[] bestBlockRoot = bestBlock.getStateRoot();
 
+            // ensure that the genesis state exists before attempting recovery
+            AionGenesis genesis = cfg.getGenesis();
+            if (!this.repository.isValidRoot(genesis.getStateRoot())) {
+                LOG.info(
+                        "Corrupt world state for genesis block hash: " + genesis.getShortHash() + ", number: " + genesis
+                                .getNumber() + ".");
+
+                buildGenesis(genesis);
```
yulongaion: `buildGenesis()` seems to be building the genesis state based on the current `repository`. It's correct for the latter usage, when the database is empty, but is it still valid here?

AlexandraRoatis: it's valid, i.e. it always results in the same state root, because when calling `createAccount(Address)` the `stateRoot = EMPTY_TRIE_HASH;`, so it does not build on the state root of the tracked repository

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -76,7 +76,12 @@ protected AionRepositoryImpl(IRepositoryConfig repoConfig) {
         // repository singleton instance
         private final static AionRepositoryImpl inst = new AionRepositoryImpl(
                 new RepositoryConfig(new File(config.getBasePath(), config.getDb().getPath()).getAbsolutePath(),
-                                     -1,
+                                     config.getDb().getPrune() > 0 ?
+                                             // if the value is smaller than backward step
```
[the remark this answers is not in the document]
AlexandraRoatis: the interaction between pruning and sync is tuned in #498

modAionImpl/src/org/aion/zero/impl/AionHub.java
```diff
@@ -272,9 +270,23 @@ private void loadBlockchain() {
...
+
+                LOG.info("Rebuilding genesis block SUCCEEDED.");
+            }
+
```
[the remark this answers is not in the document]
AlexandraRoatis: if there is anything wrong with the genesis file an error message will be printed before recovery is attempted

Issue #54, opened 2018-02-28
database pruning


AlexandraRoatis:

AlexandraRoatis:
PRs #478, #482, and #498

PR #482, opened 2018-05-14
Refactor and test journal pruning class

AlexandraRoatis:
## Description

- made the `JournalPruneDataSource.java` independent of the block and block header classes
- improved and corrected synchronization
- removed locks from `stateDatabase` since the only class accessing it is the `JournalPruneDataSource`
- further unit tests for the journal prune functionality

Related to issues #54 and #298.

## Note: this PR builds on top of #478.  

The PR in #478 should be reviewed and merged first. After which this PR can be rebased to `master-pre-merge` branch.

## Type of change

- [ ] Bug fix.
- [x] New feature.

## Testing

- full test suite and unit tests submitted as part of this PR

## Verification

### DIFF
--- a/modMcf/src/org/aion/mcf/trie/JournalPruneDataSource.java
+++ b/modMcf/src/org/aion/mcf/trie/JournalPruneDataSource.java
@@ -35,9 +35,14 @@
 package org.aion.mcf.trie;
 
 import java.util.*;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.aion.base.db.IByteArrayKeyValueDatabase;
 import org.aion.base.db.IByteArrayKeyValueStore;
 import org.aion.base.util.ByteArrayWrapper;
+import org.aion.log.AionLoggerFactory;
+import org.aion.log.LogEnum;
+import org.slf4j.Logger;
 
 /**
  * The DataSource which doesn't immediately forward delete updates (unlike inserts) but collects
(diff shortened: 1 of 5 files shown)


modMcf/src/org/aion/mcf/trie/JournalPruneDataSource.java
```diff
@@ -157,51 +213,64 @@ private Ref decRef(ByteArrayWrapper keyW) {
...
-    public synchronized void storeBlockChanges(BH header) {
-        if (!enabled) {
-            return;
+    public void storeBlockChanges(byte[] blockHash, long blockNumber) {
+        lock.writeLock().lock();
+
+        try {
+            if (!enabled) {
+                return;
+            }
```
AionJayT: should we check enable before the writelock?

modMcf/src/org/aion/mcf/trie/JournalPruneDataSource.java
```diff
@@ -85,58 +90,109 @@ public JournalPruneDataSource(IByteArrayKeyValueDatabase src) {
     }
 
     public void setPruneEnabled(boolean e) {
+        lock.writeLock().lock();
+
         enabled = e;
+
```
AionJayT: use atomicboolean for the variable - enabled, then don't need lock in this method also check enabled before the other methods before writelock.

modMcf/src/org/aion/mcf/trie/JournalPruneDataSource.java
```diff
@@ -222,29 +291,54 @@ private synchronized void rollback(BH header) {
...
     }
 
     public Optional<byte[]> get(byte[] key) {
-        return src.get(key);
+        lock.readLock().lock();
+        try {
+            return src.get(key);
+        } catch (Exception e) {
+            throw e;
+        } finally {
```
AionJayT: for the following throws(including this) , add an error log before throw if no outside classed catch these exceptions.

AlexandraRoatis:
rebased the branch to include changed to #478

qoire:
General question about the reference counts: Is it possible to remove the DB read that happens when we initialize a ref in ``incRef()``?

We only seem to use it in removing fork blocks, so I believe we can do a check if absolutely necessary in that scenario but otherwise we can ignore it.

Issue #176, opened 2018-03-13
Error handling for 


AionJayT:

AionJayT:
the method sendTransaction in the class ApiAion
if the account doesn't unlock or the user forgot the account has been locked due to the timer.
the kernel will return EMPTY_BYTE_ARRAY and get an error msg <send-transaction msg=account-not-found>. for the class ApiAion0 need to revise the behaviour if the situation happen.

PR #485, opened 2018-05-16
Fix sendTransaction error handling

aionick:
## Description

- modified ApiAion0 and ApiWeb3Aion so that they catch the exception that the sendTransaction method (in ApiAion) possibly throws (line 1211 of ApiAion0). Also modified ApiAion so that, when an exception occurs, instead of returning a 0 length byte array it propagates the exception up to the caller instead so that the exception info is not lost & masked by the byte array. Some minor typo fixes throughout ApiAion0.

Fixes Issue #176.

## Type of change

- [x] Bug fix.
- [ ] New feature.

## Testing

- Instead of figuring out how to naturally trigger an exception in sendTransaction I simply explicitly threw some exceptions at the top of the method, made calls to the api and monitored the client and server behaviour to ensure the exceptions were caught and handled as expected.

## Verification

### DIFF
--- a/modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
@@ -665,6 +665,7 @@ public class ApiWeb3Aion extends ApiAion {
             byte[] transactionHash = sendTransaction(rawTransaction);
             return new RpcMsg(TypeConverter.toJsonHex(transactionHash));
         } catch (Exception e) {
+            LOG.error("<rpc-server - exception during response for eth_sendRawTransaction: ", e);
             return new RpcMsg(null, RpcError.INTERNAL_ERROR, e.getMessage());
         }
     }
--- a/modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
@@ -661,13 +661,8 @@ public class ApiWeb3Aion extends ApiAion {
 
         byte[] rawTransaction = ByteUtil.hexStringToBytes(_rawTx);
 
-        try {
-            byte[] transactionHash = sendTransaction(rawTransaction);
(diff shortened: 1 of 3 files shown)


modApiServer/src/org/aion/api/server/http/ApiWeb3Aion.java
```diff
@@ -660,9 +660,14 @@ else if (_params instanceof JSONObject) {
             return new RpcMsg(null, RpcError.INVALID_PARAMS, "Null raw transaction provided.");
 
         byte[] rawTransaction = ByteUtil.hexStringToBytes(_rawTx);
-        byte[] transactionHash = sendTransaction(rawTransaction);
 
-        return new RpcMsg(TypeConverter.toJsonHex(transactionHash));
+        try {
```
ali-sharif: this try catch is unnecessary; 

Heuristic that I've imposed in the web3 implementation: any exceptional circumstances (ie. internal errors) in the execution of an remote procedure call should be caught by the top-level try catch:

https://github.com/aionnetwork/aion/blob/708dfee890d05f484de082c8fb6a5115d8c5ba38/modApiServer/src/org/aion/api/server/rpc/RpcProcessor.java#L80

UNLESS the exception is caused by something the user can do something about (like invalid address length, account unlocks, etc.) 

Decipline around error handling really important in languages like java, where you could end up polluting your codebase with error handling and make it more fragile and less readable by throwing ad-hoc try catches everywhere. 

@qoire - opinions?

aionick:
Also wondering if ResHandshake1Test has a problem (or is exposing an existing problem in master) or not. I brought this up with Jay, it is a randomized test and seems to fail every now and again. Tried it on William's repo and other branches the other day and got it to fail sometimes too. Might have gone undetected as such. I'm fairly confident it is not my code doing it, especially since I only failed it after pushing this last commit that simply adds a logging message.

AionJayT:
@aionick , @aion-sridevi is looking into the ResHandshake1Test issue and has a conclusion, currently solving it.

aionick:
I removed the try-catch block

PR #494, opened 2018-05-17
fix null console bug and account exist checks

aionick:
## Description

- If aion is run outside of a console environment we get a NPE. Fixes this by catching the null and recovers by prompting the user from stdin instead. Also fixed two minor instances of the same bug in Keystore, where an account address qualified as existing as long as it was a substring of a genuinely existing address.

Fixes Issue #493 .

## Type of change

- [x] Bug fix.
- [ ] New feature.

## Testing

- Can't be tested in console. Best way to test is to open debugger in IDEA with correct arguments. The affected methods in Cli class are: createAccount, exportPrivateKey, importPrivateKey. I've stepped through the debugger on all three under various circumstances.

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
+++ b/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
@@ -289,9 +289,8 @@ public class Cli {
 
         ECKey key = ECKeyFac.inst().fromPrivate(raw);
         if (key == null) {
-            System.out.println("Uable to recover private key."
-                    + "Are you sure you did not import a public key? The provided key was: "
-                    + privateKey);
+            System.out.println("Unable to recover private key."
+                    + "Are you sure you did not import a public key?");
             return false;
         }
 
--- a/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
+++ b/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
@@ -31,6 +31,9 @@
 
 package org.aion.zero.impl.cli;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import org.aion.base.util.Hex;
 import org.aion.crypto.ECKey;
 import org.aion.crypto.ECKeyFac;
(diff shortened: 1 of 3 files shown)


modMcf/src/org/aion/mcf/account/Keystore.java
```diff
@@ -231,7 +231,7 @@ public static ECKey getKey(String _address, String _password) {
         if (_address.startsWith(AION_PREFIX)) {
             List<File> files = getFiles();
             for (File file : files) {
-                if (file.getName().contains(_address)) {
+                if (file.getName().split("--")[2].equals(_address)) {
```
AionJayT: we can't assume all the files in the keystore folder are the correct naming format. Need to add more check.

modMcf/src/org/aion/mcf/account/Keystore.java
```diff
@@ -255,7 +255,7 @@ public static boolean exist(String _address) {
...
```
AionJayT: same here!

aionick:
I'll add some regex to get an exact match

modMcf/src/org/aion/mcf/account/Keystore.java
qoire: Prefer a regex match check for hexadecimal string of 64 chars (or something similar)

qoire:
Tangential, but is it possible for us to start using ``char[]`` in place of ``String`` for passwords? See here:
https://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords

aionick:
Improved the address validation; checks it is 64 bit hex and then safely uses contains.

AionJayT:
The Pattern can be a static member of the class and don't need to compile in each method call

aionick:
done

aionick:
@qoire an interesting follow-up discussion about `string` vs `char[]` I came across: https://security.stackexchange.com/questions/6753/deleting-a-java-object-securely

qoire:
@aionick Fair point, at best it provides a very minor improvement for security. Let's consider alternatives as well, perhaps moving it to a off-heap location.

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -273,9 +288,22 @@ private boolean importPrivateKey(String privateKey) {
         }
 
         ECKey key = ECKeyFac.inst().fromPrivate(raw);
+        if (key == null) {
+            System.out.println("Uable to recover private key."
```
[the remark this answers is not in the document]
aionick: fixed

PR #495, opened 2018-05-18
Redis & Redis Cluster POC

centrys-sergiu:
## Description

Introduced two new database wrappers. 
- Redis integration with batching (speed improvments)
- Redis Cluster without batching (better in mem storage) 

## Type of change

- [ ] Bug fix.
- [x] New feature.

## Testing
Compared three databases : LevelDB, RamCloud, Redis and found out that Redis would be the best option between those.

More informations in the documents here:
https://drive.google.com/open?id=1lUEyDaju_Bgj211kPhz5wElptxPRlo8E

TLDR; Redis is syncing way faster than LevelDB but the bottleneck is the memory. Redis-Cluster solves the bottleneck but sync is much slower because batching is not possible.

-

## Verification


aion-jin:
We integrate redis only when it's cluster model have better performance then LDB instead of single nodes model.
Please tested, and for those big changes please wait tech steering committee's approval before merge into master.

aion-jin:
@JayT106  don't include this POC into v0.2.8 branch.   Support 2 DB make sense, at it give user options. but support 4 DB just confusing end-users.  before we made the decision which DB should be replaced, no more DB can be added into prod binary. please do more test and give us a recommendation which one should be replaced or just keep this in a POC branch till we ready.

aion-jin:
@centrys-sergiu   please ask Jay create a aionnetwork: dev-redis,  and re-PR to poc branch first before merge into main dev.

Issue #85, opened 2018-03-02
P2pMgr - separate Tasks in other files


centrys-sergiu:
Currently TaskInbound, TaskStatus, TaskConnectPeers, TaskClear, TaskWrite are included directly in P2PMgr.

AionJayT:
close the issue, implemented by #499

PR #499, opened 2018-05-22
P2pMgr separate tasks

aionick:
## Description

- P2pMgr previously contained multiple nested classes. Each of these has been moved to its own file and a number of (mostly formatting) refactorings have been applied to improve readability and decoupling.

Fixes Issue #85 .

## Type of change

- [ ] Bug fix.
- [x] Enhancement.
- [ ] New feature.

## Testing

- I performed each refactoring in small steps and tested each refactor against our existing unit tests. I also ran the kernel with a miner and a pool and inspected the output for any abnormalities.

## Verification


aionick:
I moved the tasks into their own package. The reason I moved ChannelBuffer into the tasks package as well is because ChannelBuffer is only used by the task classes and there are a large number of references to it that would require changing some method visibilities and adding getters etc. I can put ChannelBuffer back if you guys would like.

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -577,7 +130,7 @@ public P2pMgr(int _netId, String _revision, String _nodeId, String _ip, int _por
 
         for (String _bootNode : _bootNodes) {
             Node node = Node.parseP2p(_bootNode);
-            if (node != null && validateNode(node)) {
+            if (validateNode(node)) {
```
aion-sridevi: Minor: Any reason why the null check is removed? If the first condition is false, validateNode is not called (&& is short circuited).

aionick: I did this because validateNode immediately returns false when its input is null, seemed to me like an unnecessary duplication of a null check

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -992,79 +154,75 @@ public void run() {
...
     public void register(final List<Handler> _cbs) {
         for (Handler _cb : _cbs) {
             Header h = _cb.getHeader();
             short ver = h.getVer();
             byte ctrl = h.getCtrl();
             if (Ver.filter(ver) != Ver.UNKNOWN && Ctrl.filter(ctrl) != Ctrl.UNKNOWN) {
-                if (!versions.contains(ver)) {
-                    versions.add(ver);
-                }
+                versions.add(ver);
```
aion-sridevi: Is this change intended?

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
aion-sridevi: I believe validateNode will not be called when node is null.
 https://stackoverflow.com/questions/8759868/java-logical-operator-short-circuiting

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
aionick: since versions is a set the add call won't allow duplicates

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
aionick: That’s true, sorry I didn’t mean the check is duplicated, I meant the validate function handles the null case already so the explicit null check before it isn’t needed for correctness. But you’re right no need to enter the function and grow the call stack etc.

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskSend.java
```diff
@@ -0,0 +1,171 @@
...
+         * @param _dest
+         */
+        public MsgOut(int _nodeId, String _displayId, Msg _msg, Dest _dest) {
+            nodeId = _nodeId;
+            displayId = _displayId;
+            msg = _msg;
+            dest = _dest;
+            timestamp = System.currentTimeMillis();
+        }
+    }
```
AionJayT: Should separate this inner class out since many tasks use it.

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskReceive.java
```diff
@@ -0,0 +1,101 @@
...
+         * @param route
+         * @param msg
+         */
+        public MsgIn(int nid, String nsid, int route, byte[] msg) {
+            this.nid = nid;
+            this.nsid = nsid;
+            this.route = route;
+            this.msg = msg;
+        }
+    }
```
AionJayT: Should separate this inner class out since many tasks use it.

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskInbound.java
```diff
@@ -0,0 +1,578 @@
...
+import org.aion.p2p.impl.zero.msg.ResHandshake;
+import org.aion.p2p.impl.zero.msg.ResHandshake1;
+import org.aion.p2p.impl1.P2pException;
+import org.aion.p2p.impl1.P2pMgr;
+import org.aion.p2p.impl1.P2pMgr.Dest;
+import org.aion.p2p.impl1.tasks.TaskReceive.MsgIn;
+import org.aion.p2p.impl1.tasks.TaskSend.MsgOut;
+
+public class TaskInbound implements Runnable {
+    private final P2pMgr mgr;
```
AionJayT: better to use Interface class.

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskInbound.java
```diff
@@ -0,0 +1,578 @@
...
+    private final Selector selector;
+    private final NodeMgr nodeMgr;
```
AionJayT: better to use Interface class

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskConnectPeers.java
```diff
@@ -0,0 +1,181 @@
...
+import org.aion.p2p.impl.zero.msg.ReqHandshake1;
+import org.aion.p2p.impl1.P2pMgr;
+import org.aion.p2p.impl1.P2pMgr.Dest;
+import org.aion.p2p.impl1.tasks.TaskSend.MsgOut;
+
+public class TaskConnectPeers implements Runnable {
+    private static final int PERIOD_CONNECT_OUTBOUND = 1000;
+    private static final int TIMEOUT_OUTBOUND_CONNECT = 10000;
+
+    private final NodeMgr nodeMgr;
```
AionJayT: better to use Interface class

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskConnectPeers.java
```diff
@@ -0,0 +1,181 @@
...
+    private final int maxActiveNodes;
+    private final P2pMgr mgr;
+    private AtomicBoolean start;
+    private LinkedBlockingQueue<MsgOut> sendMsgQue;
+    private Selector selector;
+    private ReqHandshake1 cachedReqHandshake1;
```
AionJayT: better to use the base class or create an interface class

PR #501, opened 2018-05-23
Dev redis q2

centrys-sergiu:
## Description

Introduced two new database wrappers. 
- Redis integration with batching (speed improvments)
- Redis Cluster without batching (better in mem storage) 

## Type of change

- [ ] Bug fix.
- [x] New feature.

## Testing
Compared three databases : LevelDB, RamCloud, Redis and found out that Redis would be the best option between those.

More informations in the documents here:
https://drive.google.com/open?id=1lUEyDaju_Bgj211kPhz5wElptxPRlo8E

TLDR; Redis is syncing way faster than LevelDB but the bottleneck is the memory. Redis-Cluster solves the bottleneck but sync is much slower because batching is not possible.

-

## Verification


modDbImpl/src/org/aion/db/impl/redisdb/RedisClusterWrapper.java
```diff
@@ -0,0 +1,221 @@
...
+            }
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+        }
+    }
+
+    @Override
+    public void close() {
+        try {
+            db.close();
```
AlexandraRoatis: the `isOpen()` call would still return `true` after this call; perhaps the db object should be set to null and we can check for null in `isOpen()`

PR #537, opened 2018-06-26
Initial version of modGui

aion-kelvin:
## Description

Initial version of Aion kernel GUI.  Supports a single use case: open GUI, launch the kernel, exit GUI, re-open GUI, terminate kernel.

Dependencies
 - aion_api: needed small change to postpack.sh script.  It is pushed to remote branch but not yet merged into dev of aion_api: https://github.com/aionnetwork/aion_api/commits/dev-modGui

Note on class/code design:
```
The GUI application code uses a variation of MVC.  For us, we'll divide up the                                                                                                                                      
responsibilities into the following layers:

1) Model: any class that interfaces with components outside of the GUI; i.e. KernelLauncher
   for interface with the OS to launch processes; KernelConnection provides interface to
   Aion API.  The state of Aion API lives in the model layer.  The model does not know about
   Controller or View, it just publishes events into an EventBus.

2) View: the layer that contains the UI elements; i.e. all of the JavaFX .fxml files.  The 
   elements in the View (i.e. a Button) have associated handler methods, which are part of
   Controller.  Through these associated methods, View knows about Controller and 
   delegates control to classes/methods in Controller.  The View is very "dumb" and doesn't
   have logic to modify itself.

3) Controller: this layer mediates interaction between Model and View by serving two purposes:
     (a) Contains handlers to which the View is registered (through the view's FXML).
         Usually these handlers manipulate Model through some method call.  The heavy
         lifting of the logic should live in Model.
     (b) Subscribe to events published by Model and update View accordingly.  Events that it
         may need to handle include kernel state change (e.g. # of peers).  EventBus is the 
         mechanism used to facilitate PubSub pattern.
   Generally, each View component has one or zero accompanying classes that are part of the 
   Controller layer (usually a class name ending with "Controller").  Within it are the 
   handlers for its corresponding View and also handlers for Model events that would cause a
   change in that View component.

Depending on how the code develops we can revisit this.  The fact that Controller layer has 
two purposes may end up getting unwieldy; if so, maybe move (a) into the responsibility of
View or call it out as something else ("View controller"?).
```

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [x] Unit test.

## Testing

- 58% lines coverage for unit and integ tests.  Most of the uncovered lines come from classes that will likely change when developing the next feature.  Jenkins output: https://ci.aion.network:8443/job/aionnetwork/job/dev-modGui2/1/
- unit tests for most Model classes
- integ tests for most Controller classes and accompanying Views
- manual testing of GUI

## Verification

- [ ] I have made appropriate changes to the corresponding documentation. [there is none currently]


modGui/src/org/aion/os/KernelLauncher.java
```diff
@@ -0,0 +1,226 @@
...
+        FileOutputStream fos = new FileOutputStream(pidFile);
+        ObjectOutputStream oos = new ObjectOutputStream(fos);
+        oos.writeObject(kernel);
+        setCurrentInstance(kernel);
+        return kernel;
+    }
+
+    @VisibleForTesting void removePersistedPid() throws IOException {
+        if(pidFile.exists() && pidFile.isFile()) {
+            pidFile.delete();
```
JayT106: should catch the exception when deleting the file.

qoire:
Is this the final version of the first phase of modGui? @aion-kelvin @AionJayT

modMcf/src/org/aion/mcf/config/CfgGuiLauncher.java
```diff
@@ -0,0 +1,125 @@
...
+                        default:
+                            break;
+                    }
+                    break;
+                case XMLStreamReader.END_ELEMENT:
+                    break loop;
+            }
+        }
+    }
+
```
JayT106: Do you need some settings (or default settings) been wrote into the config.xml? if the answer is yes, should update the config.xml in the modBoot/resource.

modGui/src/org/aion/gui/controller/HeaderPaneControls.java
```diff
@@ -0,0 +1,150 @@
...
+            } else {
+                styleClass.add(STYLE_DEFAULT);
+                setStyleToChildren(headerButton, "header-button-label");
+            }
+        }
+    }
+
+    private void setStyleToChildren(Node headerButton, String styleClassToSet) {
+        VBox vbox = (VBox) ((VBox) headerButton).getChildren().get(0);
+        ObservableList<String> styleClass = vbox.getChildren().get(0).getStyleClass();
```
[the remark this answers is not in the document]
aion-kelvin: In this case yes, because the headerButton parameter is intended to be one of the buttons in the top navigation pane of the GUI.  The buttons are always a VBox (a rectangular box) and then has a Label as child (the text of the button).  Nothing prevents someone from using that setStyleToChildren method wrong though -- they could call with the headerButton parameter being something else, possibly with 0 children.

modGui/src/org/aion/gui/model/KernelConnection.java
```diff
@@ -0,0 +1,135 @@
...
+        // currently it is restricted and the recommended way to call the API is to subclass
+        // from AbstractAionApiClient which provides subclasses with synchronized blocks to
+        // call their critical sections with.
+        return this.api;
+    }
+
+    private String getConnectionString() {
+        final String protocol = "tcp";
+        final String ip = Preconditions.checkNotNull(cfgApi.getZmq().getIp());
+        final String port = Preconditions.checkNotNull(String.valueOf(cfgApi.getZmq().getPort()));
```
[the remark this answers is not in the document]
aion-kelvin: If the Preconditions.checkNotNull fails (and throws NPE), there is nothing we can really do to recover.  So that's why I did not put a catch.

I guess we can catch it and present a nicer error via GUI instead of the program just crashing and exiting, but this is something I'm planning to add later.

Let me know if you still want me to add a catch.

modGui/src/org/aion/os/KernelLauncher.java
aion-kelvin: Ok, will change

modMcf/src/org/aion/mcf/config/CfgGuiLauncher.java
aion-kelvin: Ok, will update

modBoot/resource/config.xml
```diff
@@ -84,5 +84,18 @@
...
+			<!--Whether JVM settings for launching kernel should be autodetected; 'true' or 'false'-->
+			<autodetect>true</autodetect>
+			<!--Path to JAVA_HOME.  This field has no effect if autodetect is true.-->
+			<java-home>aion.sh</java-home>
+			<!--Working directory of kernel process.  This field has no effect if autodetect is true.-->
+			<working-dir>/placeholder/for/aion_root_dir</working-dir>
+			<!--Filename of aion launcher script, relative to working-dir.  This field has no effect if autodetect is true.-->
+			<aion-sh></aion-sh>
+		</launcher>
+	</gui>
```
AlexandraRoatis: I don't currently see the advantage of having this configuration? Perhaps it should be hidden.

modGui/integration-test/org/aion/gui/controller/AbstractControllerTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+import static org.mockito.ArgumentMatchers.isNotNull;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class AbstractControllerTest {
```
AlexandraRoatis: This class should not be added in this state. It can be left for a later PR.

modMcf/test/org/aion/mcf/config/CfgLogTest.java
```diff
@@ -223,4 +223,9 @@ public FileVisitResult postVisitDirectory(final Path dir, final IOException e) t
 
         return true;
     }
+
+    @Test
+    public void test() {
+        System.out.println( new CfgLog().toXML() );
```
AlexandraRoatis: probably shouldn't be included in the PR

script/nohup_wrapper.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# TODO redirect to sane location
+nohup $* > /tmp/aion-kernel-nohup &
```
AlexandraRoatis: this should be placed in the local folder, otherwise, the user may never become aware of it

AlexandraRoatis: Ok. That sounds good.

PR #540, opened 2018-07-03
changed getStorageValue to return an optional

aionick:
## Description

- Changed IRepositoryQuery so that its getStorageValue return type is an Optional. This forces implementors to handle the 'missing value' case identically, and this consistency is important for some pre-compiled contracts. This PR current has the .gitmodules pointing to the add_optional2db branch in the fastVM. Once  <a href="https://github.com/aionnetwork/aion_fastvm/pull/12">this corresponding PR</a> in the fastVM is merged I will update the .gitmodules file to no longer point to this branch, it's just here in the meantime for Jenkins.

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- Ran the kernel and tested against the current test suite.

## Verification

### DIFF
--- a/modAionImpl/test/org/aion/zero/impl/db/AionRepositoryImplTest.java
+++ b/modAionImpl/test/org/aion/zero/impl/db/AionRepositoryImplTest.java
@@ -1,4 +1,4 @@
-/* ******************************************************************************
+/*
  * Copyright (c) 2017-2018 Aion foundation.
  *
  *     This file is part of the aion network project.
@@ -31,7 +31,7 @@
  *     Samuel Neves through the BLAKE2 implementation.
  *     Zcash project team.
  *     Bitcoinj team.
- ******************************************************************************/
+ */
 package org.aion.zero.impl.db;
 
 import static com.google.common.truth.Truth.assertThat;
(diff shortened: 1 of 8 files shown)


modAionImpl/test/org/aion/zero/impl/db/AionRepositoryImplTest.java
```diff
@@ -146,7 +146,7 @@ public void testAccountStateUpdateStorageRow() {
         track.flush();
 
         byte[] retrievedValue =
-                repository.getStorageValue(defaultAccount, new DataWord(key)).getNoLeadZeroesData();
+                repository.getStorageValue(defaultAccount, new DataWord(key)).get().getNoLeadZeroesData();
```
JayT106: what if the getStorageValue return Optional.empty() ?

aionick: I'll update these tests so that if empty is returned they fail

qoire:
Not sure I see the benefits of adding ``Optional`` into our critical path. Invoking an additional object on every storage query is sure to add some overhead. Why not just document that storageValue can possibly return null?

I understand that ``Optional`` would enforce this relationship, but probably not a luxury we can afford on our critical path.

AlexandraRoatis:
@qoire do you still see it as a problem if the implementation is changed to pass up the object created in the `IKeyValueStore`?

PR #541, opened 2018-07-03
Small optimizations

AlexandraRoatis:
## Notice

## Description

* ignoring blocks received with null peer status, i.e. not requested;
* removed redundant p2p variable from TaskImportBlocks;
* removed direct reference of databases from AionBlockStore; avoiding redundant repeated flush on same objects.

Fixes Issue # N/A.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing unit tests

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
@@ -58,9 +58,7 @@ public class AionBlockStore extends AbstractPowBlockstore<AionBlock, A0BlockHead
 
     protected ReadWriteLock lock = new ReentrantReadWriteLock();
 
-    private IByteArrayKeyValueDatabase indexDS;
     private DataSourceArray<List<BlockInfo>> index;
-    private IByteArrayKeyValueDatabase blocksDS;
     private ObjectDataSource<AionBlock> blocks;
 
     private boolean checkIntegrity = true;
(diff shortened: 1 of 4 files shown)


modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -1178,12 +1159,12 @@ public void close() {
         lock.writeLock().lock();
 
         try {
-            indexDS.close();
+            index.close();
         } catch (Exception e) {
```
aion-kelvin: does index.close() actually throw Exception?  it looks like it just throws IOException

modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -1178,12 +1159,12 @@ public void close() {
...
             LOG.error("Not able to close the index database:", e);
         } finally {
             try {
-                blocksDS.close();
+                blocks.close();
             } catch (Exception e) {
```
aion-kelvin: does index.close() actually throw Exception?  it looks like it just throws IOException

modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -1178,12 +1159,12 @@ public void close() {
...
 
         try {
-            indexDS.close();
+            index.close();
         } catch (Exception e) {
             LOG.error("Not able to close the index database:", e);
         } finally {
             try {
-                blocksDS.close();
+                blocks.close();
```
aion-kelvin: Does anything bad happen if index failed to close, but blocks gets closed?

AlexandraRoatis: If either one fails to close it may lead to data corruption, but most of the time nothing bad happens. There are methods in place to recover from data corruption. In a bad case, the user may need to revert back a few blocks. In the worst case, they may need to wipe the database. There is functionality for correcting the index database, so that might be recovered. If the block database is broken, they need to re-sync from Genesis.

PR #571, opened 2018-07-31
[RLP 1] Module testing and clean-up

AlexandraRoatis:
## Description

- added unit tests based on [rlptest.json](https://github.com/ethereum/tests/blob/develop/RLPTests/rlptest.json) and additional examples I considered relevant
- removed some unused & untested functionality from `RLP.java`
- optimized the `CompactEncoder.java` functionality to reduce the number of created byte arrays
- some other edge case unit tests

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

- existing test suite and added unit tests
- also cherry-picked these changes to the master branch and ran the kernel

## Verification


modRlp/src/org/aion/rlp/CompactEncoder.java
```diff
@@ -112,16 +110,35 @@
...
+                ret[0] = (byte) flag;
+                ret[1] = 0;
+                System.arraycopy(nibbles, 0, ret, 2, nibbles.length);
+            } else { // terminator == 1
+                ret = new byte[nibbles.length + 1];
+                ret[0] = (byte) flag;
+                ret[1] = 0;
+                System.arraycopy(nibbles, 0, ret, 2, nibbles.length - 1);
+            }
+            nibbles = ret;
```
AionJayT: This method seems like can be refactored cause the nibbles been copied twice during the logic (including the following ByteArrayOutputStream write). Considering find a way to avoid too many copies in this method.

modRlp/build.xml
```diff
@@ -16,6 +16,9 @@
   <path id="classpath.test">
     <pathelement location="${dir.lib}/junit_4/junit-4.12.jar"/>
     <pathelement location="${dir.lib}/hamcrest/hamcrest-all-1.3.jar"/>
+    <pathelement location="${dir.lib}/truth-0.42.jar"/>
+    <pathelement location="${dir.lib}/guava-25.1-jre.jar"/>
+    <pathelement location="${dir.lib}/diffutils-1.3.0.jar"/>
```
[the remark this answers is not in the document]
AlexandraRoatis: I think we should. The versions we are using are quite outdated.

modRlp/src/org/aion/rlp/CompactEncoder.java
AlexandraRoatis: optimized in PR #572 commit 50ce43b

modRlp/src/org/aion/rlp/RLP.java
```diff
@@ -647,23 +458,24 @@ public static DecodeResult decode(byte[] data, int pos) {
...
-                                                                        // its
-                                                                        // own
-                                                                        // RLP
-                                                                        // encoding
-        } else if (prefix < OFFSET_LONG_ITEM) {
+            return new DecodeResult(pos + 1, new byte[] {data[pos]}); // byte
+            // is
+            // its
+            // own
+            // RLP
```
aion-kelvin: awkward

modRlp/src/org/aion/rlp/Utils.java
```diff
@@ -66,10 +79,14 @@
...
     static byte[] hexEncode(byte[] in) {
+        return hexEncode(in, false);
+    }
+
+    static byte[] hexEncode(byte[] in, boolean withTerminatorByte) {
         if (in == null) {
             return null;
         }
-        byte[] ret = new byte[in.length * 2];
+        byte[] ret = new byte[in.length * 2 + (withTerminatorByte ? 1 : 0)];
```
aion-kelvin: Why do we add that extra byte at the end that never gets populated?  (Don't think the loop below ever writes into ret[in.length*2+1])

modRlp/test/org/aion/rlp/UtilsTest.java
```diff
@@ -0,0 +1,83 @@
...
+    public void testHexEncode_wSingleByte() {
+        for (byte b : Utils.encodingTable) {
+            byte[] input = new byte[] {b};
+            assertThat(Utils.hexEncode(input)).isEqualTo(Hex.encode(input));
+        }
+    }
+
+    @Test
+    public void testHexEncode_wString() {
+        String value = "1234567890abcdef";
```
aion-kelvin: Not directly related to your code, but I was experimenting this with test and I added an 'a' to this string, so value = "1234567890abcdefa" and the Hex.decode(value) throws an exception.  Is that expected?

modRlp/test/org/aion/rlp/RLPTest.java
```diff
@@ -95,41 +89,18 @@
...
-        byte[] payload = Hex.decode(peersPacket);
-
-        byte[] ip = decodeIP4Bytes(payload, 5);
-
-        assertEquals(InetAddress.getByAddress(ip).toString(), ("/54.204.10.41"));
-    }
-
-    @Test
-    public void test2() throws UnknownHostException {
+    public void test2() {
```
aion-kelvin: Optional: maybe rename this test() or test1()

modRlp/test/org/aion/rlp/RLPSpecExtraTest.java
```diff
@@ -0,0 +1,372 @@
...
+import org.aion.base.util.Hex;
+import org.junit.Test;
+
+/**
+ * Tests for the RLP implementation that are complementary to RLP specification tests from Ethereum
+ * implemented in {@link RLPSpecTest}.
+ *
+ * @author Alexandra Roatis
+ */
+public class RLPSpecExtraTest {
```
aion-kelvin: Looks good thanks for the extra tests, this looks pretty well covered.

One question -- I didn't see any numerical tests where the input is a negative number.  Is that supposed to be possible?  To encode a negative number, then decode the result, and get back the same negative number?

modRlp/src/org/aion/rlp/CompactEncoder.java
```diff
@@ -31,56 +31,54 @@
...
  *
- * An alternative way of thinking about this to not think of there being a
- * terminator symbol, but instead treat bit specifying the existence of the
- * terminator symbol as a bit specifying that the given node encodes a final
- * node, where the value is an actual value, rather than the hash of yet another
- * node.
+ * <p>To solve both of these issues, we force the first nibble of the final byte-stream to encode
+ * two flags, specifying oddness of length (ignoring the 'T' symbol) and terminator status; these
+ * are placed, respectively, into the two lowest significant bits of the first nibble. In the case
+ * of an even-length hex string, we must introduce a second nibble (of value zero) to ensure the
```
arajasek: I understand what it's saying, but I find this last sentence quite confusingly worded. If it's not a bother, I'd suggest rephrasing it so that it doesn't sound like "we must ensure that even-length hex strings are even-length hex strings".

modRlp/src/org/aion/rlp/CompactEncoder.java
AionJayT: good job!

modRlp/src/org/aion/rlp/RLP.java
```diff
@@ -396,47 +193,60 @@ public static void fullTraverse(byte[] msgData, int level, int startPos, int end
                     int length = calcLength(lengthOfLength, msgData, pos);
 
                     // now we can parse an item for data[1]..data[length]
-                    System.out.println("-- level: [" + level + "] Found big list length: " + length);
+                    System.out.println(
+                            "-- level: [" + level + "] Found big list length: " + length);
```
AionJayT: Does this be printed to the terminal during kernel running? If yes, can we add new logEnum for the RLP module?

modRlp/test/org/aion/rlp/RLPSpecExtraTest.java
```diff
@@ -0,0 +1,372 @@
...
+
+    // ENCODING
+
+    @Test
+    public void testEncodeShort3() {
```
aionick: Can we also test the max value for `short` and the min values for `short`,`int`,`long` (looks like our min is zero?) just to have those edges covered

modRlp/test/org/aion/rlp/UtilsTest.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+package org.aion.rlp;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.util.Arrays;
+import org.aion.base.util.Hex;
+import org.junit.Test;
+
+public class UtilsTest {
```
aionick: Could we also have tests for concatenating two empty arrays as well as empty with non-empty arrays?

modRlp/src/org/aion/rlp/RLP.java
AlexandraRoatis: I removed this functionality in #572 because it wasn't used.

modRlp/src/org/aion/rlp/Utils.java
AlexandraRoatis: I did this as an optimization to avoid creating and copying several byte arrays. This way the proper length array gets created from the start. Since this is just a utility, I prefer keeping it as is.

modRlp/test/org/aion/rlp/DecodeResultTest.java
```diff
@@ -0,0 +1,34 @@
...
+ *     If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Contributors:
+ *     Aion foundation.
+ */
+package org.aion.rlp;
+
+import org.junit.Test;
+
+public class DecodeResultTest {
```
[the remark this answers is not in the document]
AlexandraRoatis: The class is actually very well covered because of the `RLP.decode` calls.

modRlp/test/org/aion/rlp/RLPElementTest.java
```diff
@@ -0,0 +1,20 @@
...
+import static com.google.common.truth.Truth.assertThat;
+
+import org.aion.base.util.ByteUtil;
+import org.junit.Test;
+
+public class RLPElementTest {
+
+    @Test(expected = RuntimeException.class)
+    public void testRLPList_wRuntimeException() {
+        RLPList.recursivePrint(null);
```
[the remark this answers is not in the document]
AlexandraRoatis: The class is very well covered because of the `RLP.decode2` calls. Only the null checks were missing, and it felt more elegant to put both tests in one class instead of having a class for each with a single test. If you insist I can split them into individual classes. Do you? :)

modRlp/test/org/aion/rlp/RLPElementTest.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    @Test
+    public void testRLPItem_wNull() {
+        RLPItem item = new RLPItem(null);
```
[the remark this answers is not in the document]
AlexandraRoatis: see comment above

modRlp/test/org/aion/rlp/RLPSpecExtraTest.java
AlexandraRoatis: Nope, the encoding does not work with negative numbers. They get interpreted as positive, so you would get a different value when decoding.

modRlp/test/org/aion/rlp/RLPSpecExtraTest.java
AlexandraRoatis: That's covered in the `RLPSpecTest.java` which is based on [rlptest.json](https://github.com/ethereum/tests/blob/develop/RLPTests/rlptest.json). These tests complement those test cases with values that I considered may be relevant.

modRlp/test/org/aion/rlp/UtilsTest.java
AlexandraRoatis: `Hex.decode(value)` works only with even length strings because one byte is encoded to two characters.

modRlp/test/org/aion/rlp/RLPElementTest.java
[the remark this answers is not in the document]
aion-kelvin: Oh, I see, makes sense... we can leave it this way for now then.

Philosophically, for a class "X" like having a corresponding "XTest.java," but in this case I think it really doesn't make much difference.

modRlp/test/org/aion/rlp/RLPElementTest.java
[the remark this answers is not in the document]
aion-kelvin: Sounds good

modRlp/src/org/aion/rlp/CompactEncoder.java
```diff
@@ -112,16 +110,35 @@
 
         if (nibbles[nibbles.length - 1] == TERMINATOR) {
```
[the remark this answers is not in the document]
AlexandraRoatis: done in PR #572 commit ba5c2954639c95eee8258f5d9c4392d1ad147225

modRlp/test/org/aion/rlp/CompactEncoderTest.java
```diff
@@ -31,93 +31,115 @@
...
-import org.junit.Test;
-
+import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.assertArrayEquals;
 
-import org.aion.rlp.CompactEncoder;
+import java.util.Arrays;
+import org.junit.Test;
 
 public class CompactEncoderTest {
```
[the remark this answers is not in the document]
AlexandraRoatis: done in PR #572 commit ba5c2954639c95eee8258f5d9c4392d1ad147225

modRlp/src/org/aion/rlp/RLP.java
AlexandraRoatis: fixed in PR #572 commit ba5c2954639c95eee8258f5d9c4392d1ad147225

modRlp/test/org/aion/rlp/RLPTest.java
AlexandraRoatis: gave it a meaningful name in PR #572 commit ba5c2954639c95eee8258f5d9c4392d1ad147225

modRlp/test/org/aion/rlp/UtilsTest.java
AlexandraRoatis: done in PR #572 commit ba5c2954639c95eee8258f5d9c4392d1ad147225

modRlp/src/org/aion/rlp/CompactEncoder.java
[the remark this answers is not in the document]
AlexandraRoatis: improved in PR #572 commit ba5c2954639c95eee8258f5d9c4392d1ad147225

PR #573, opened 2018-08-01
Add java-side unit tests to fastVM

aionick:
## Description

- Adds java unit tests to the major classes related to the fastVM. Some refactorings for testing were done and some for cleanups. The cleanups were only done in small classes like ExecutionContext/Helper/Result and not the larger ones, where with lack of existing unit tests I didn't feel confident enough to alter much.
- Currently this PR uses branch `test_fastvm` in the fastvm until that branch is merged into `precompiled` in <a href="https://github.com/aionnetwork/aion_fastvm/pull/15">this PR</a>.

Fixes Issue # N/A.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [x] Unit test.

## Testing

- The tests I added!

## Verification


modAion/src/org/aion/zero/types/AionTxExecSummary.java
```diff
@@ -340,6 +340,7 @@ public BigInteger getFee() {
         if (!parsed) {
             rlpParse();
         }
+        // redundant getReceipt just calls itself??
```
AionJayT: Can't understand this comment.

modAionImpl/src/org/aion/zero/impl/AionGenesis.java
```diff
@@ -46,7 +46,7 @@
 public class AionGenesis extends AionBlock {
     /**
      * Make this value constant for now, in the future we may move this to a
-     * more configuration position, this indicates the address at which the
+     * more configuration position, this indicates the getRecipient at which the
```
AionJayT: looks like the refactor influence the "address"

modAionImpl/src/org/aion/zero/impl/A0BCConfig.java
```diff
@@ -33,7 +33,7 @@
      * Retrieve the currently set coinbase for this particular node, blocks
      * mined with this node will use this as the coinbase.
      *
-     * @return {@code coinbase} a 32-bytes address
+     * @return {@code coinbase} a 32-bytes getRecipient
```
AionJayT: looks like the refactor influence the "address"

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -1273,7 +1273,7 @@ public boolean isBlockExist(byte[] hash) {
     }
 
     /**
-     * Finds up to limit blocks starting from blockNumber on main chain
+     * Finds up to limit blocks starting from getBlockNumber on main chain
```
AionJayT: looks like the refactor influence the "blockNumber"

modVM/build.xml
```diff
@@ -40,55 +41,12 @@
   </target>
 
   <target name="test_build" depends="build">
```
AionJayT: why delete the contain of the ant task

modVM/src/org/aion/vm/AbstractExecutor.java
```diff
@@ -93,13 +97,29 @@ protected ITxExecSummary execute(ITransaction tx, long contextNrgLmit) {
...
+     */
+    protected final boolean prepare(ITransaction tx, long contextNrgLmit) {
         if (isLocalCall) {
             return true;
         }
 
-        // check nrg limit
-        BigInteger txNrgPrice = BigInteger.valueOf(tx.getNrgPrice());
+        // check nrg limit      use signum 1 for txNrgPrice?
+        BigInteger txNrgPrice = BigInteger.valueOf(tx.getNrgPrice()).abs();
```
AionJayT: the pendingstateImpl will check the nrgPrice can not be a negative value.

modVM/src/org/aion/vm/ExecutionContext.java
```diff
@@ -31,277 +31,254 @@
  * @author yulong
  */
 public class ExecutionContext {
-
+    private static final String NEG_MSG = "must be non-negative.";
```
AionJayT: what's the NEG means?

modVM/src/org/aion/vm/ExecutionContext.java
```diff
@@ -31,277 +31,254 @@
...
+
+    /**
+     * Returns the length of the big-endian binary encoding of this ExecutionContext.
+     *
+     * @return the legtn of this ExecutionContext's binary encoding.
+     */
+    private int getEncodingLength() {
+        return (Address.ADDRESS_LEN * 4) + (DataWord.BYTES * 3) + (Long.BYTES * 4) +
+            (Integer.BYTES * 4) + callData.length;
+    }
```
AionJayT: can make (Address.ADDRESS_LEN * 4) + (DataWord.BYTES * 3) + (Long.BYTES * 4) +
            (Integer.BYTES * 4) as a static method, or a static int declaration

modVM/src/org/aion/vm/ExecutionHelper.java
```diff
@@ -63,126 +69,78 @@ public Call(byte[] data, byte[] destination, byte[] value) {
...
+     * Adds a collection of internal transactions, txs, to the internal transactions list.
      *
-     * @param txs
+     * @param txs The collection of internal transactions to add.
      */
     public void addInternalTransactions(Collection<AionInternalTx> txs) {
-        internalTxs.addAll(txs);
+        for (AionInternalTx tx : txs) {
+            if (tx != null) { this.internalTxs.add(tx); }
+        }
```
AionJayT: can null AionInternalTx happens in the Collection<AionInternalTx>?

modVM/src/org/aion/vm/TransactionResult.java
```diff
@@ -81,6 +83,23 @@ public TransactionResult() {
...
+     *
+     * @return a big-endian binary encoding of this TransactionResult.
+     */
+    public final byte[] toBytes() {
+        ByteBuffer buffer = ByteBuffer.allocate(4 + 8 + 4 + (output == null ? 0 : output.length));
+        buffer.order(ByteOrder.BIG_ENDIAN);
+        buffer.putInt(code.toInt());
+        buffer.putLong(nrgLeft);
+        buffer.putInt(output == null ? 0 : output.length);
+        if (output != null) {
```
AionJayT: The behavior is different compared with the ExecutionResult.toBytes(), any reason? if the behavior can be the same, can we put this method to the abstract class?

aionick: Yes the two are the same. At first it seemed inappropriate to me to produce sub-class encodings in a superclass, since the children may acquire fields or other changes over time that would demand their own encodings. I suppose though the better choice is to have a generic toBytes in the super and then override when necessary, so I'll put that back as it was

modVM/build.xml
aionick: I put the lines back but CTRL+/ commented them out. modVM doesn't have tests yet and ant complains when it goes looking for them

aionick:
The reason I added the line `this.output = (output == null) ? ByteUtil.EMPTY_BYTE_ARRAY : output;` to AbstractExecutionResult is to fix a small bug where, if output is null, once you encode the object using toBytes and then decode using parse, output always gets set to an empty byte array and is never set null, so you don't get the same object back truly. Other code depends on this empty byte usage so instead I just made sure the constructor gets rid of the null right away

aionick:
Also: TransactionResult was not a unit test priority for me atm, the vm is mostly interested in ExecutionResult. The classes are similar and TransactionResult is mostly simple 1-line methods, I will add testing for this class as part of my next PR

modVM/src/org/aion/vm/ExecutionHelper.java
AionJayT: cause I think what happens in the internal txs is once an internal tx failed, then stop the following tx executing and rollback to the previous status. That's why I think should not happen a null AionInternalTx in the collection. @yulongnuco Can you confirm my words?

modVM/src/org/aion/vm/AbstractExecutor.java
```diff
@@ -49,8 +48,12 @@
     private long blockRemainingNrg;
     private boolean askNonce = true;
 
-    public AbstractExecutor(IRepository _repo,
-        boolean _localCall, long _blkRemainingNrg, Logger _logger) {
+    protected void setExecutionResult(IExecutionResult result) {
```
aion-kelvin: optional: think it makes more sense for this to be after the constructor

modVM/src/org/aion/vm/ExecutionContext.java
```diff
@@ -31,277 +31,254 @@
...
+    public void setAddress(Address address) {
+        this.address = address;
+    }
+
+    /**
+     * Sets the transaction hash to txHash.
+     *
+     * @param txHash The new transaction hash.
+     */
+    public void setTransactionHash(byte[] txHash) { this.txHash = txHash; }
```
aion-kelvin: newline

modVM/src/org/aion/vm/ExecutionHelper.java
```diff
@@ -63,126 +69,78 @@ public Call(byte[] data, byte[] destination, byte[] value) {
...
+     * Adds the collection addresses to the set of deleted accounts if addresses is non-null.
      *
-     * @param addresses
+     * @param addresses The addressed to add to the set of deleted accounts.
      */
     public void addDeleteAccounts(Collection<Address> addresses) {
-        for (Address address : addresses) {
-            addDeleteAccount(address);
+        for (Address addr : addresses) {
+            if (addr != null) { deleteAccounts.add(addr); }
```
aion-kelvin: nit/optional: I would have the if check in addDeleteAccount(Address) and then this loop just calls addDeleteAccount(Address)... not a big deal though

modVM/src/org/aion/vm/ExecutionHelper.java
```diff
@@ -203,4 +161,33 @@ public void merge(ExecutionHelper other, boolean success) {
...
+     * @return the deleted accounts list.
+     */
+    public List<Address> getDeleteAccounts() { return new ArrayList<>(deleteAccounts); }
+
+    /**
+     * Returns the execution logs.
+     *
+     * @return the execution logs.
+     */
+    public List<Log> getLogs() { return logs; }
```
aion-kelvin: Maybe overly-paranoid, but would it be worth returning Collections.unmodifiableList(logs) instead?  That wrapper prevents someone from getting the logs and then modifying it, affecting the behaviour in this object

modVM/src/org/aion/vm/ExecutionHelper.java
```diff
@@ -203,4 +161,33 @@ public void merge(ExecutionHelper other, boolean success) {
...
+
+    /**
+     * Returns the calls.
+     *
+     * @return the calls.
+     */
+    public List<Call> getCalls() { return calls; }
```
aion-kelvin: same comment as above

modVM/src/org/aion/vm/ExecutionHelper.java
```diff
@@ -203,4 +161,33 @@ public void merge(ExecutionHelper other, boolean success) {
...
+
+    /**
+     * Returns the internal transactions.
+     *
+     * @return the internal transactions.
+     */
+    public List<AionInternalTx> getInternalTransactions() { return internalTxs; }
```
aion-kelvin: same comment as above

modVM/src/org/aion/vm/TransactionResult.java
```diff
@@ -22,6 +22,8 @@
  */
 package org.aion.vm;
 
+import java.nio.ByteBuffer;
```
aion-kelvin: these don't seem necessary

modVM/src/org/aion/vm/TransactionExecutor.java
```diff
@@ -50,7 +51,7 @@
  * @author yulong
  */
 public class TransactionExecutor extends AbstractExecutor {
-
+    // provider is essential to execute, but it is only set manually with a setter, not good practice
```
aion-kelvin: should we just add it as parameter to the constructor?  (if it takes too long to refactor the existing calls to this constructor, we just change all the existing calls to the constructor to pass in null, that's should end up being the exact same behaviour)

modVM/src/org/aion/vm/TransactionExecutor.java
```diff
@@ -177,7 +178,7 @@ protected void call() {
     }
 
     /**
-     * Prepares contract create.
+     * Prepares contract create.    is prepares the right word here?
```
aion-kelvin: I don't know

modVM/src/org/aion/vm/Forks.java
```diff
@@ -0,0 +1,22 @@
+package org.aion.vm;
+
+public class Forks {
+
+    private static final long SEPTEMBER_FORK = 1000000L;
```
aion-kelvin: Maybe name it SEPTEMBER_2018 fork... in case we have another fork after that, that's also in September :)

modPrecompiled/build.xml
```diff
@@ -127,6 +129,8 @@
             <batchtest todir="${test.report.dir}">
                 <fileset dir="${src.test}">
                     <include name="**/*Test.java"/>
+                    <exclude name="**/AionAuctionContractTest.java"/>
+                    <exclude name="**/TokenBridgeContractTest.java"/>
```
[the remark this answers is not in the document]
aionick: These were leftover from the branch I branched off because they needed some fixes, thanks

modVM/src/org/aion/vm/ExecutionContext.java
aionick: what's the issue here?

modVM/src/org/aion/vm/ExecutionHelper.java
aionick: I agree, I'll check with yulong about this tomorrow, the null checks may not even be necessary

modVM/src/org/aion/vm/TransactionExecutor.java
aionick: maybe this is almost worth a separate PR. I'm not sure what the motivation behind this was, it may have a use. I think at least we could instantiate the provider so that if someone forgets to explicitly set it nothing crashes

modVM/src/org/aion/vm/AbstractExecutionResult.java
```diff
@@ -119,79 +147,65 @@ public long getNrgLeft() {
...
      *
-     * @param code The new code.
-     * @param nrgLeft The new nrgLeft.
+     * @param code the integer representation of the code to be set.
      */
-    public void setCodeAndNrgLeft(int code, long nrgLeft) {
-        setCode(code);
+    public void setCode(int code) {
+        ResultCode resCode = ResultCode.fromInt(code);
+        this.code = resCode;
```
AlexandraRoatis: the resCode variable is not actually needed

modVM/src/org/aion/vm/ExecutionContext.java
aion-kelvin: It looks awkward to not have a newline between the left curly brace { and the method body; and between right curly brace } and the method body.

All the other methods are formatted that way; i.e. setAddress, getEncodingLength.

modVM/src/org/aion/vm/TransactionExecutor.java
aion-kelvin: Sounds fair ... it's not really a big deal for me, but if not fixed now, it'll probably be a while before anyone else notices it again :)

modVM/src/org/aion/vm/AbstractExecutor.java
```diff
@@ -93,13 +92,28 @@ protected ITxExecSummary execute(ITransaction tx, long contextNrgLmit) {
...
+     * @return true if call is local or if all criteria listed above are met.
+     */
+    protected final boolean prepare(ITransaction tx, long contextNrgLmit) {
         if (isLocalCall) {
             return true;
         }
 
-        // check nrg limit
-        BigInteger txNrgPrice = BigInteger.valueOf(tx.getNrgPrice());
+        BigInteger txNrgPrice = BigInteger.valueOf(tx.getNrgPrice()).abs();
```
AlexandraRoatis: When can the value of `tx.getNrgPrice()` be negative? And by making it positive can we end up with incorrect results?

modVM/src/org/aion/vm/AbstractExecutor.java
```diff
@@ -217,4 +258,9 @@ protected void updateRepo(ITxExecSummary summary, ITransaction tx,
             LOGGER.debug("Transaction logs: {}", summary.getLogs());
         }
     }
+
+    protected void setExecutionResult(IExecutionResult result) {
```
AlexandraRoatis: not used

modVM/src/org/aion/vm/AbstractExecutor.java
AionJayT: there is no negative value happen due to the txValidator set a rule to check it.

AionJayT: and agree should not manipulate the arg in tx

modVM/src/org/aion/vm/ExecutionContext.java
aionick: sorry intellij tucks the one-line methods like that and I couldn't see the difference

modVM/src/org/aion/vm/AbstractExecutionResult.java
aionick: thanks, I had intended to check for null first

modVM/src/org/aion/vm/AbstractExecutor.java
aionick: it's used by the FastVM in the `fastvm_test` branch, you might have been on the `precompiled` branch

modVM/src/org/aion/vm/ExecutionHelper.java
```diff
@@ -122,11 +122,13 @@ public void addCall(byte[] data, byte[] destination, byte[] value) {
     }
 
     /**
-     * Adds an internal transaction, tx, to the internal transactions list.
+     * Adds an internal transaction, tx, to the internal transactions list.cd
```
AlexandraRoatis: accidental addition I believe

PR #582, opened 2018-08-07
Incorporate wallet into modGui

aion-kelvin:
## Notice

## Description

Basically pulled in most of the functionality from aion_ui repo, refactored it to improve testability, and incorporated it into the Aion Kernel GUI.  Added a lot of unit tests (most of the code of the Model layer) and some integration tests (mostly around interaction between Controller and View).

Still in progress of adding more integ tests, but wanted to send out this PR.  Not expecting much refactoring of the existing code while I add more integ tests.

Features that should work with this PR:
- Launch/terminate kernel process
- Create HD wallet account
- Import/export of account
- Recovery of wallet from mnemonic
- Send transactions
- Receive transactions (just a helpful screen that displays your current account in QR code form)
- Transaction history
- Editing of config.xml (Kernel must not be running)
- Had to limit the size of the window, account page's list view can't seem to resize properly

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- Unit tests for most code in Model layer
- Integ tests for about half of the UI interactions (Dashboard, Accounts, Settings, Receive screens have complete integ tests; Send, and Transaction History screens still need more integ tests)
- Manual testing for main use cases (launch/terminate kernel, create accounts, recover accounts, import/export, tx history)

## Verification

- [x] I have added tests for my fix or feature. (mostly)


aion-kelvin:
@AlexandraRoatis 



> The aion_gui.sh start script does not get packed with the build. I think it should be.

I will look into this.

> I observed some strange behavior wrt loading account. After creating an account and then adding a few more, if I unlock the seed account only the first account is loaded. Is this by design? Do the accounts need to have a balance to load?`

I think it is by design; I will double check and let you know (or fix it if it needs to be fixed)

>    After integrating with the CLI changes, it would be nice to see which network we are connected to mainnet or conquest.

For now it will just support mainnet.  Would it suffice to just put some text into the documentation and/or readme to say that GUI can only connect you to mainnet at the moment?

aionick:
gui
I'm getting this this error when I try and terminate the kernel, it just hangs and I have to `CTRL+C` to fully quit. Seemed to be the result of messing with the config while the kernel was running (I deleted all the text in the config the first time, next time just a bracket). Not sure if this is local to me or not

aionick:
Just a bit of "pressure testing" type stuff. I launched kernel and then sent a transaction as quickly as I could and while it was sending I deleted half the text in the config and hit apply & save and now the `seding transaction...` is hanging. Eventually the generate transaction button greyed out, probably when the account was unactive for too long, and then the kernel threw this same exception over and over:
`Exception in thread "api" java.lang.NullPointerException
	at org.aion.gui.controller.SendController.lambda$refreshAccountBalance$7(SendController.java:296)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$9(PlatformImpl.java:418)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:417)
	at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.gtk.GtkApplication.lambda$runLoop$11(GtkApplication.java:277)
	at java.base/java.lang.Thread.run(Thread.java:844)
`

modGui/src/org/aion/gui/model/TransactionProcessor.java
```diff
@@ -0,0 +1,270 @@
...
+        final BlockDTO oldestSafeBlock = accountManager.getOldestSafeBlock(addresses, nullSafeBlockFilter);
+        if (oldestSafeBlock != null) {
+            processTransactionsFromBlock(oldestSafeBlock, addresses);
+        }
+    }
+
+    private static int getCores() {
+        int cores = Runtime.getRuntime().availableProcessors();
+        if (cores > 1) {
+            cores = cores / 2;
```
AionJayT: this is not true, cause not every cpu support HT, or the HT been diabled in the BIOS.

modGui/src/org/aion/wallet/crypto/MasterKey.java
```diff
@@ -0,0 +1,45 @@
...
+        for (final int pathElement : derivationPath) {
+            key = getChild(pathElement, key);
+        }
+        final byte[] seed = Arrays.copyOfRange(key, 0, 32);
+        return new SeededECKeyEd25519(seed);
+
+    }
+
+    private byte[] getChild(final int pathElement, final byte[] keyHash) throws ValidationException {
+        byte[] parentPrivateKey = Arrays.copyOfRange(keyHash, 0, 32);
```
AionJayT: should we catch the ArrayIndexOutOfBoundsException exception in case
key length change in the future?

modGui/src/org/aion/gui/model/ConsoleTail.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Create a user-friendly string representing a time interval.  The string will be as follows:
+     *
+     * if interval is larger than 1 day: "YYYY-mm-dd HH:mm AM/PM"
+     * if interval is between 1 hour and 1 day: "today HH:mm AM/PM"
+     * if interval is between 1 minute and 1 hour: "x minutes ago"
+     * if interval is between 1 second and 1 minute: "less than a minute ago"
+     *
+     * @param start
+     * @param end
```
aionick: Can remove these params

modGui/src/org/aion/gui/model/TransactionProcessor.java
```diff
@@ -0,0 +1,270 @@
...
+        for (String address : addresses) {
+            final List<TransactionDTO> txs = new ArrayList<TransactionDTO>(accountManager.getTransactions(address));
+            final Iterator<TransactionDTO> iterator = txs.iterator();
+            final List<TransactionDTO> oldTxs = new ArrayList<>();
+            while (iterator.hasNext()) {
+                final TransactionDTO t = iterator.next();
+                if (t.getBlockNumber() > previousSafe) {
+                    oldTxs.add(t);
+                } else {
+                    continue;
```
aionick: this can be removed

modGui/src/org/aion/gui/model/ConsoleTail.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public class ConsoleTail {
+    private final ScheduledExecutorService timer;
```
aionick: unused

aionick:
-I noticed there were a few unused methods in various classes, are you mostly hanging on to these because you expect them to be used in the future?
-The `ApiType` class is completely unused it looks like.
-The Close button in the import window is almost the same colour as the background, maybe worth using a more distinguishing colour?

modGui/src/org/aion/wallet/util/AddressUtils.java
```diff
@@ -0,0 +1,21 @@
...
+        return address != null && !address.isEmpty() && isAionAddress(address);
+    }
+
+    public static boolean equals(final String addrOne, final String addrTwo) {
+        return TypeConverter.toJsonHex(addrOne).equals(TypeConverter.toJsonHex(addrTwo));
+    }
+
+    private static boolean isAionAddress(final String address) {
+        final boolean isFull = address.startsWith("0xa") && address.length() == 66;
+        final boolean isStripped = address.startsWith("a") && address.length() == 64;
```
AionJayT: in the current definition, the address should start with "0xa0" or "a0", also make sure the input string is lowercase

aion-kelvin:
@aionick 

> Just a bit of "pressure testing" type stuff. I launched kernel and then sent a transaction as quickly as I could and while it was sending I deleted half the text in the config and hit apply & save and now the seding transaction... is hanging. Eventually the generate transaction button greyed out, probably when the account was unactive for too long, and then the kernel threw this same exception over and over [...]

That's a bit concerning.  I'll try to reproduce and figure out if there's some easy fix.  Most likely there's some unreliable assumptions about ordering of events and threads, that causes it to go into some messed up state when violated...

> Also interesting -- the account I was sending the transaction too, another account in my wallet, has now disappeared after quitting and restarting the gui. After hitting add account I re-generated a new account that has the 5 aion I transferred. I can't confirm whether this new account has the same address as the one that disappeared.

@AlexandraRoatis also mentioned this.  It is a bug -- it's not saving accounts properly beyond the first created one.  I'll have this fixed in my next PR update.  The new account would have had the same address, since any account beyond the first one uses a priv key that is derived from the priv key of that first account in a deterministic way.

modGui/src/org/aion/gui/model/ConsoleTail.java
aion-kelvin: Will fix

modGui/src/org/aion/gui/model/TransactionProcessor.java
aion-kelvin: Ok

modGui/src/org/aion/gui/model/ConsoleTail.java
aion-kelvin: Oops, in constructor I accidentally use the reference from the constructor arguments, instead of the member variable.  Will fix that.

modGui/build.xml
```diff
@@ -10,16 +10,25 @@
...
+	<property name="native.dir" value="native" />
 
     <!-- Dependencies for building production code (i.e. non-test files) -->
     <path id="classpath.dependency">
         <pathelement location="${dir.mod}/modLogger.jar"/>
         <pathelement location="${dir.mod}/modMcf.jar"/>
         <pathelement location="${dir.mod}/modAionImpl.jar"/>
+        <pathelement location="${dir.mod}/modCrypto.jar"/>
+        <pathelement location="${dir.mod}/modAionBase.jar"/>
         <pathelement location="${dir.lib}/../aion_api/pack/modAionApi.jar" />
```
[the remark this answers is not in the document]
aion-kelvin: Since aion_api is its own Git repo, it seems strange for its build.xml to the aion/mod folder.  I thought it was better that aion_api's build.xml not assume that it is being called from aion's build.xml.

For now, modified aion's build.xml to copy from aion_api/pack/modAionApi.jar to the mod directory, and modGui's build.xml references that.  Let me know if you see any problems with that.

modGui/src/org/aion/gui/model/BalanceRetriever.java
```diff
@@ -0,0 +1,36 @@
...
+     * Get balance of given account
+     *
+     * @param address address of account
+     * @return if API is connected, balance of that account; otherwise, null
+     */
+    public BigInteger getBalance(String address) {
+        // TODO Prefer Optional over null.
+        final BigInteger balance;
+        if(!apiIsConnected()) {
+            balance = null;
```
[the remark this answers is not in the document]
aion-kelvin: I double checked that all the places that this is called has the null check.

modGui/src/org/aion/gui/model/TransactionProcessor.java
aion-kelvin: I'm not totally sure why this was coded this way when I inherited the code.  I think here, "getCores" is supposed to mean "get the number of cores this Transaction Processor should use," not "get the number of cores this machine has."

(This value gets passed into a thread-pool executor; it wouldn't make sense to use every core of the machine in a single executor)

modGui/src/org/aion/wallet/crypto/MasterKey.java
aion-kelvin: Ok

modGui/src/org/aion/wallet/util/AddressUtils.java
aion-kelvin: Added lowercase restriction and a few test cases for this method

aion-kelvin:
@aionick 

> I noticed there were a few unused methods in various classes, are you mostly hanging on to these because you expect them to be used in the future?

It's probably a combination of things that will be used in the future and code that I borrowed from aion_ui that didn't end up getting used but was not removed.  I will run 'find unused declarations' in the IDE to try to get rid of them.

> The ApiType class is completely unused it looks like.

It's not doing anything useful at the moment, but LightAppSettings code is expecting it to be passed in at some point and then it gets saved to a file.  Kind of silly but I don't want to do the refactor+verification for it at the moment, if this issue isn't too pressing.

> The Close button in the import window is almost the same colour as the background, maybe worth using a more distinguishing colour?

I will see if I can make a quick fix for this

aion-kelvin:
@aionick 
>I noticed there were a few unused methods in various classes, are you mostly hanging on to these because you expect them to be used in the future?

I looked at the unused methods.  A lot of the ones IntelliJ complains about is because they get called implicitly by EventBus (feature of Guava) and method calls bound to the FXML that JavaFX will implicitly call.  There was also some methods that aren't currently used, but makes sense for a class to have, i.e. "getXXX()" -- might be useful when adding unit tests later.  I think I've removed most of the remaining ones that are not used.  Let me know if you spot any other ones that should be deleted.

modGui/src/org/aion/gui/model/BalanceRetriever.java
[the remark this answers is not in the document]
AionJayT: cool

PR #585, opened 2018-08-08
ModApiServer unit tests

arajasek:
## Description

This PR adds unit tests for the ApiServer module, primarily for the classes Api, APiAion, and ApiAion0.

Also included is a change to ApiAion0.process(). In the f_compile_VALUE case, we should be checking for "compile-error", since that is the error flag introduced by Api.ContractCreateResult.contract_compileSolidity().

Api.java is modified to include 2 new constructors - one for standard purposes, and one only for testing.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [x] Unit test.

## Testing

Adds the first tests for the ApiServer module, passes the Jenkins build.

## Verification

### DIFF
--- a/modApiServer/src/org/aion/api/server/Api.java
+++ b/modApiServer/src/org/aion/api/server/Api.java
@@ -50,7 +50,20 @@ public abstract class Api<B extends AbstractBlock<?, ?>> {
 
     private final AccountManager ACCOUNT_MANAGER = AccountManager.inst();
     private final Compiler solc = Compiler.getInstance();
-    protected final AionPendingStateImpl pendingState = AionPendingStateImpl.inst();
+    protected final AionPendingStateImpl pendingState;
+
+    // This is the constructor that should always be used, unless testing
+    Api() {
+        pendingState = AionPendingStateImpl.inst();
+    }
+
+    // Only for testing purposes
+    Api(boolean test) {
+        if (!test)
+            pendingState = AionPendingStateImpl.inst();
+        else
+            pendingState = null;
+    }
 
     public abstract String getCoinbase();
 
--- a/modApiServer/src/org/aion/api/server/pb/ApiAion0.java
+++ b/modApiServer/src/org/aion/api/server/pb/ApiAion0.java
@@ -665,7 +665,7 @@ public class ApiAion0 extends ApiAion implements IApiAion {
                         Message.rsp_compile.Builder b = Message.rsp_compile.newBuilder();
 
                         for (Entry<String, CompiledContr> entry : _contrs.entrySet()) {
-                            if (entry.getKey().contains("AionCompileError")) {
+                            if (entry.getKey().contains("compile-error")) {
                                 byte[] retHeader =
                                     ApiUtil.toReturnHeader(
                                         getApiVersion(),
--- a/modApiServer/test/org/aion/api/server/ApiAionTests.java
+++ b/modApiServer/test/org/aion/api/server/ApiAionTests.java
@@ -2,7 +2,6 @@ package org.aion.api.server;
 
 import org.aion.api.server.types.ArgTxCall;
 import org.aion.api.server.types.SyncInfo;
-import org.aion.api.server.types.TxRecpt;
 import org.aion.base.type.Address;
 import org.aion.base.type.ITransaction;
 import org.aion.base.type.ITxReceipt;
--- a/modApiServer/test/org/aion/api/server/ApiAionTests.java
+++ b/modApiServer/test/org/aion/api/server/ApiAionTests.java
@@ -1,6 +1,8 @@
 package org.aion.api.server;
 
+import org.aion.api.server.types.ArgTxCall;
 import org.aion.api.server.types.SyncInfo;
+import org.aion.api.server.types.TxRecpt;
 import org.aion.base.type.Address;
 import org.aion.base.type.ITransaction;
 import org.aion.base.type.ITxReceipt;
@@ -8,6 +10,8 @@ import org.aion.crypto.ed25519.ECKeyEd25519;
 import org.aion.evtmgr.impl.evt.EventBlock;
 import org.aion.evtmgr.impl.evt.EventDummy;
 import org.aion.evtmgr.impl.evt.EventTx;
+import org.aion.mcf.account.AccountManager;
+import org.aion.mcf.account.Keystore;
 import org.aion.zero.impl.blockchain.AionImpl;
 import org.aion.zero.impl.config.CfgAion;
 import org.aion.zero.impl.db.AionBlockStore;
@@ -100,7 +104,6 @@ public class ApiAionTests {
         ApiAionTests.ApiAionImpl api = new ApiAionTests.ApiAionImpl(impl);
         System.out.println("API Version = " + api.getApiVersion());
         assertNotNull(api.getInstalledFltrs());
-        assertNotNull(api.getCoinbase());
     }
 
     @Test
@@ -109,10 +112,10 @@ public class ApiAionTests {
         AionImpl impl = AionImpl.inst();
         ApiAionTests.ApiAionImpl api = new ApiAionTests.ApiAionImpl(impl);
         api.initNrgOracle(impl);
-        assertNotNull(ApiAion.NRG_ORACLE);
+        assertNotNull(api.getNrgOracle());
         // Initing a second time should not create a new NrgOracle
         api.initNrgOracle(impl);
-        assertNotNull(ApiAion.NRG_ORACLE);
+        assertNotNull(api.getNrgOracle());
     }
 
     @Test
@@ -180,10 +183,10 @@ public class ApiAionTests {
         ApiAionTests.ApiAionImpl api = new ApiAionTests.ApiAionImpl(impl);
 
         AionBlock parentBlk = impl.getBlockchain().getBestBlock();
-        String msg = "test message";
+        byte[] msg = "test message".getBytes();
         AionTransaction tx = new AionTransaction(repo.getNonce(Address.ZERO_ADDRESS()).toByteArray(),
                 Address.ZERO_ADDRESS(), Address.ZERO_ADDRESS(), BigInteger.ONE.toByteArray(),
-                    msg.getBytes(),100000, 100000);
+                    msg,100000, 100000);
         tx.sign(new ECKeyEd25519());
 
         AionBlock blk = impl.getAionHub().getBlockchain().createNewBlock(parentBlk,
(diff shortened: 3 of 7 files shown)


modApiServer/test/org/aion/api/server/ApiAionTests.java
```diff
@@ -0,0 +1,398 @@
...
+
+            //provokes exception in EpApi.run()
+            ees.add(new EventBlock(EventBlock.CALLBACK.ONBLOCK0));
+            ees.add(new EventDummy());
+        }
+
+    }
+
+    @Test
+    public void TestCreate() {
```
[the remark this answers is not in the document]
arajasek: Thanks! Do you know whether we have style guidelines anywhere? I followed the notation in modAionApi's BaseAPITests class, which uses CamelCase and prints out the name each time. It would be good if we have some "official" guidelines.

modApiServer/src/org/aion/api/server/Api.java
```diff
@@ -50,7 +50,20 @@
...
-    protected final AionPendingStateImpl pendingState = AionPendingStateImpl.inst();
+    protected final AionPendingStateImpl pendingState;
+
+    // This is the constructor that should always be used, unless testing
+    Api() {
+        pendingState = AionPendingStateImpl.inst();
+    }
+
+    // Only for testing purposes
+    Api(boolean test) {
```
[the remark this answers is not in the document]
AionJayT: @VisibleForTesting require guava lib to support, please include the guava lib as a dependency

modApiServer/src/org/aion/api/server/pb/ApiAion0.java
```diff
@@ -665,7 +665,7 @@ public ApiAion0(IAionChain ac) {
                         Message.rsp_compile.Builder b = Message.rsp_compile.newBuilder();
 
                         for (Entry<String, CompiledContr> entry : _contrs.entrySet()) {
-                            if (entry.getKey().contains("AionCompileError")) {
+                            if (entry.getKey().contains("compile-error")) {
```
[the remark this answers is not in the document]
AionJayT: @arajasek can you check with @yulongaion? I believe it has been changed due to the solidity compiler change.

modApiServer/test/org/aion/api/server/ApiAionTests.java
```diff
@@ -0,0 +1,398 @@
...
+    @Test
+    public void TestDoCall() {
+        System.out.println("run TestDoCall.");
+        AionImpl impl = AionImpl.inst();
+        AionRepositoryImpl repo = AionRepositoryImpl.inst();
+        ApiAionTests.ApiAionImpl api = new ApiAionTests.ApiAionImpl(impl);
+
+        byte[] msg = "test message".getBytes();
+
+        Address addr = new Address(Keystore.create("testPwd"));
```
AionJayT: should clean the keystore file after test

modApiServer/src/org/aion/api/server/Api.java
AionJayT: Also, my p2p PR update the mockito version, you can just mock the final class without changing the code. please considering to remove the pendingState.

modApiServer/test/org/aion/api/server/ApiAionTests.java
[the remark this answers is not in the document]
AlexandraRoatis: We try to follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html). Older code has not been brought up to date yet which is unfortunatly the case with the test cases you looked at.

modApiServer/test/org/aion/api/server/ApiAionTests.java
```diff
@@ -0,0 +1,398 @@
...
+        {
+            Thread.sleep(5000);
+        }
+        catch(InterruptedException ex) {}
+        api.shutDownES();
+        assertTrue(api.allFlagsSet());
+    }
+
+    @Test
+    public void TestGetBlock() {
```
aionick: Probably not worth splitting into 3 separate tests since the cases are so small, but it would be valuable to break up the 3 cases with comments -> using -1, 0 and then n > 0

modApiServer/test/org/aion/api/server/ApiAionTests.java
```diff
@@ -0,0 +1,398 @@
...
+        AionImpl impl = AionImpl.inst();
+        ApiAionTests.ApiAionImpl api = new ApiAionTests.ApiAionImpl(impl);
+        System.out.println("API Version = " + api.getApiVersion());
+        assertNotNull(api.getInstalledFltrs());
+    }
+
+    @Test
+    public void TestInitNrgOracle() {
+        System.out.println("run TestInitNrgOracle.");
+        AionImpl impl = AionImpl.inst();
```
aionick: I would pull this line (or if you have any other ones shared by every test) out into a `setup` method. With JUnit you can use the `@Before` annotation to tell it to run that method prior to every test invocation. (`@After` to do any cleanup)

modApiServer/test/org/aion/api/server/ApiAionTests.java
```diff
@@ -0,0 +1,398 @@
...
+    @Test
+    public void TestStartES() {
+        System.out.println("run TestStartES.");
+        AionImpl impl = AionImpl.inst();
+        ApiAionTests.ApiAionImpl api = new ApiAionTests.ApiAionImpl(impl);
+        api.startES("thName");
+        api.addEvents();
+        try
+        {
+            Thread.sleep(5000);
```
aionick: Can we avoid sleeping here? Or at least for 5 seconds? Also, if the exception is thrown would it cause your assert to fail? If so, a print message or something could be useful otherwise the failure will appear very cryptic

aionick:
I'm only partially finished reviewing, will get back to this tomorrow!

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
```diff
@@ -0,0 +1,2137 @@
...
+        api.shutDown();
+    }
+
+    @Test
+    public void TestProcessProtocolVersion() {
+        System.out.println("run TestProcessProtocolVersion.");
+        AionImpl impl = AionImpl.inst();
+        ApiAion0 api = new ApiAion0(impl);
+
+        byte[] request = ByteBuffer.allocate(msg.length + REQ_HEADER_NOHASH_LEN + hash.length).put(api.getApiVersion())
```
aionick: Looks like this block of code repeats enough in other tests that it's worth extracting into a helper method

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
```diff
@@ -0,0 +1,2137 @@
...
+
+        assertEquals(Message.Retcode.r_success_VALUE, rsp[1]);
+
+        try {
+            Message.rsp_protocolVersion pv = Message.rsp_protocolVersion.parseFrom(stripHeader(rsp));
+            assertEquals(Integer.toString(api.getApiVersion()), pv.getApi());
+            assertEquals(Version.KERNEL_VERSION, pv.getKernel());
+            assertEquals(EquihashMiner.VERSION, pv.getMiner());
+        }
+        catch (Exception e) {
```
aionick: very minor - typical java style places the `catch` on the same line as the `}`
(also `fail()` can print a message so you could put the `e.getMessage()` inside the `fail` :) )

I don't think these `try-catch` blocks are too necessary though since what they do is print a fail message and then fail. If you don't use these blocks the exception will be thrown and cause the test to fail anyway - plus we will get a stack trace instead of just the error message - which is more handy for debugging

modApiServer/src/org/aion/api/server/pb/ApiAion0.java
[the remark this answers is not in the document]
arajasek: Will do!

modApiServer/test/org/aion/api/server/ApiAionTests.java
aion-kelvin: It's not clear to me why the sleep is needed.  Could you explain that?

I agree about the InterruptedException... actually, I'd probably just delete that catch and add 'throws Exception' to the method signature

modApiServer/test/org/aion/api/server/ApiAionTests.java
```diff
@@ -0,0 +1,398 @@
...
+        assertEquals(blk, api.getBestBlock());
+
+        assertEquals(blk, api.getBlock(-1));
+
+        assertEquals(blk.toString(), api.getBlockByHash(blk.getHash()).toString());
+        assertEquals(blk.toString(), api.getBlock(blk.getNumber()).toString());
+
+        Map.Entry rslt = api.getBlockWithTotalDifficulty(blk.getNumber());
+        assertEquals(rslt.getKey().toString(), blk.toString());
+        if (!blk.isGenesis())
```
aion-kelvin: How come this if block is needed?  Shouldn't you be able to know whether blk.isGenesis() is expected to be true or false?

modApiServer/test/org/aion/api/server/ApiAionTests.java
```diff
@@ -0,0 +1,398 @@
...
+
+    @Test
+    public void TestGetSync() {
+        System.out.println("run TestGetSync.");
+        AionImpl impl = AionImpl.inst();
+        ApiAionTests.ApiAionImpl api = new ApiAionTests.ApiAionImpl(impl);
+        SyncInfo sync = api.getSync();
+        assertNotNull(sync);
+        assertEquals(sync.done, impl.isSyncComplete());
+        if (impl.getInitialStartingBlockNumber().isPresent())
```
aion-kelvin: Same question about the if statements

modApiServer/test/org/aion/api/server/ApiAionTests.java
aion-kelvin: Indeed

modApiServer/test/org/aion/api/server/ApiUtilTests.java
```diff
@@ -0,0 +1,160 @@
...
+        hash = RandomUtils.nextBytes(ApiUtil.HASH_LEN);
+        error = RandomUtils.nextBytes(errorLength);
+        result = RandomUtils.nextBytes(resultLength);
+    }
+
+    @Test
+    public void TestNulls() {
+        System.out.println("run TestNulls.");
+
+        assertNull(ApiUtil.toReturnHeader(vers, retCode, null));
```
aion-kelvin: Could you print out the variables that are randomized, so if this fails in CI server, we can reproduce it?

You might have to do that in every test method because IIRC, I think JUnit reconstructs the test class for each test method

modApiServer/test/org/aion/api/server/ApiUtilTests.java
```diff
@@ -0,0 +1,160 @@
...
+    public void TestHeaderHashErrorResult() {
+        System.out.println("run TestHeaderHashErrorResult.");
+
+        byte[] header = ApiUtil.toReturnHeader(vers, retCode, hash, error, result);
+        assertEquals(vers, header[0]);
+        assertEquals(retCode, header[1]);
+        assertEquals(1, header[2]);
+        assertArrayEquals(hash, Arrays.copyOfRange(header, 3, ApiUtil.HASH_LEN + 3));
+        assertEquals(error.length, header[ApiUtil.HASH_LEN + 3]);
+        assertArrayEquals(error, Arrays.copyOfRange(header, ApiUtil.HASH_LEN + 4, ApiUtil.HASH_LEN + errorLength + 4));
```
aion-kelvin: Out of curiosity, how did you determine these constants?  i.e. + 4, + 3, etc

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
aion-kelvin: +1 for abandoning the catch block in favor or just saying 'throws Exception' in method signature

modApiServer/test/org/aion/api/server/ApiUtilTests.java
```diff
@@ -0,0 +1,160 @@
...
+import org.apache.commons.lang3.RandomUtils;
+import org.junit.Test;
+
+import java.util.Arrays;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class ApiUtilTests {
```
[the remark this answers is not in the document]
arajasek: Will do! The tests in modAionApi don't follow this...Is this something we should look into?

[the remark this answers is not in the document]
aionick: Ah that's because those tests don't get run by `ant`. I believe because they require the kernel to be running in order to work and even then there's tests that rely on certain accounts being in your keystore etc.

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
```diff
@@ -0,0 +1,1304 @@
...
+    }
+
+    private void tearDown() {
+        api.shutDown();
+        rsp = null;
+    }
+
+    @Test
+    public void testHeartBeatMsg() {
+        setup();
```
aionick: With JUnit you can actually just put a `@Before` annotation above your setup method and `@After` above your tearDown and they will be called before and after every single test case respectively, and then you can remove the explicit calls to setup and teardown :)

modApiServer/test/org/aion/api/server/ApiUtilTests.java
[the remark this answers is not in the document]
arajasek: done

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
```diff
@@ -0,0 +1,2137 @@
...
+
+        if (hasHash) {
+            return Arrays.copyOfRange(rsp, RSP_HEADER_LEN, RSP_HEADER_LEN + bodyLen);
+        } else {
+            return Arrays.copyOfRange(rsp, RSP_HEADER_NOHASH_LEN, RSP_HEADER_NOHASH_LEN + bodyLen);
+        }
+    }
+
+    @Test
+    public void TestCreate() {
```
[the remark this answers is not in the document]
arajasek: this test was fairly meaningless...i've deleted it

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
```diff
@@ -0,0 +1,2137 @@
...
+
+        byte[] msg = ByteBuffer.allocate(api.getApiHeaderLen()).put(api.getApiVersion())
+                .put((byte) Message.Servs.s_hb_VALUE).array();
+        assertTrue(ApiAion0.heartBeatMsg(msg));
+        assertFalse(ApiAion0.heartBeatMsg(null));
+
+        msg[0] = 0;
+        assertFalse(ApiAion0.heartBeatMsg(msg));
+
+        api.shutDown();
```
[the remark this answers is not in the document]
arajasek: done

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
arajasek: done

modApiServer/test/org/aion/api/server/pb/ApiAion0Tests.java
arajasek: done

modApiServer/test/org/aion/api/server/ApiAionTests.java
arajasek: so we have a race-condition here...the test only passes if the events dispatched in addEvents() are executed before we pass onto the allFlagsSet assertion...My hacky way of ensuring that always happens is by waiting 5 seconds (which is almost definitely excessive).

what's the "correct" way to enforce this? use a lock of some sort?

ps: sorry i'm weak at concurrency-related stuff :P

modApiServer/test/org/aion/api/server/ApiUtilTests.java
arajasek: done

aionick:
approved, pending others

modApiServer/test/org/aion/api/server/ApiAionTests.java
arajasek: done

modApiServer/test/org/aion/api/server/ApiUtilTests.java
arajasek: I reverse-calculated them from the functions in ApiUtil.java, which let me figure out exactly what the contents of the headers are.

modApiServer/src/org/aion/api/server/Api.java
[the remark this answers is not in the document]
arajasek: done

arajasek: i wasn't able to mock the AionPendingStateImpl.inst() because it's a static method, and mockito doesn't seem to support mocking statics.

modApiServer/src/org/aion/api/server/pb/ApiAion0.java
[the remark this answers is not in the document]
arajasek: @yulongaion confirms that this is good :)

modApiServer/test/org/aion/api/server/ApiAionTests.java
arajasek: they're optionals, so it just felt like good practice to check that there's a value present before poking it

modApiServer/test/org/aion/api/server/ApiAionTests.java
arajasek: i don't understand when the bestBlock is the genesis block...but it doesn't seem necessary, so i've gotten rid of the if block

aion-kelvin: Sounds good

PR #589, opened 2018-08-09
APB - 46 Precompiled contract to allow ED25519 signature verification

centrys-david:
## Description

This adds a new precompiled contract that validates a if a message was signed by a specific account.
```
     * input
     * |
     * | message hash (32 bytes) | public key (32 bytes) | signature bytes (64 bytes) |
     * <p>
     * where
     * - message hash = keccak 256 hashed message
     * - public key = public key of the ED25519 key pair
     * - signature = 64 bytes of the ED25519 signature
```
**Important**: There is **NO solidity** support for this functionality to be mapped to a function due to the limitations of the `aion_fastvm`.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

* `EDVerifyContractTest.java`

## Verification

### DIFF
--- a/modPrecompiled/src/org/aion/precompiled/ContractFactory.java
+++ b/modPrecompiled/src/org/aion/precompiled/ContractFactory.java
@@ -28,6 +28,7 @@ import org.aion.base.vm.IDataWord;
 import org.aion.mcf.core.AccountState;
 import org.aion.mcf.db.IBlockStoreBase;
 import org.aion.precompiled.contracts.ATB.TokenBridgeContract;
+import org.aion.precompiled.contracts.EDVerifyContract;
 import org.aion.vm.ExecutionContext;
 import org.aion.vm.IContractFactory;
 import org.aion.vm.IPrecompiledContract;
@@ -43,7 +44,9 @@ public class ContractFactory implements IContractFactory {
     private static final String TOKEN_BRIDGE = "0000000000000000000000000000000000000000000000000000000000000200";
     private static final String TOKEN_BRIDGE_INITIAL_OWNER = "a008d7b29e8d1f4bfab428adce89dc219c4714b2c6bf3fd1131b688f9ad804aa";
 
-    public ContractFactory(){}
+    private static final String ED_VERIFY = "0000000000000000000000000000000000000000000000000000000000000010";
+
+    private ContractFactory(){}
 
     /**
      * Returns a new pre-compiled contract such that the address of the new contract is address.
@@ -68,6 +71,8 @@ public class ContractFactory implements IContractFactory {
                     return null;
                 
                 return contract;
+            case ED_VERIFY:
+                return new EDVerifyContract();
             default: return null;
         }
     }
@@ -94,5 +99,12 @@ public class ContractFactory implements IContractFactory {
     public static Address getTotalCurrencyContractAddress() {
         return Address.wrap(TOTAL_CURRENCY);
     }
+
+    /**
+     * Returns the address of the EdVerifyContract contract.
+     *
+     * @return the contract address
+     */
+    public static Address getEdVerifyContractAddress() { return Address.wrap(ED_VERIFY); }
 
 }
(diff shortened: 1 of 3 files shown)


modPrecompiled/src/org/aion/precompiled/contracts/EDVerifyContract.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public IExecutionResult execute(byte[] input, long nrgLimit) {
+        if (COST > nrgLimit) {
+            return new ExecutionResult(ExecutionResult.ResultCode.OUT_OF_NRG, 0, Address.ZERO_ADDRESS().toBytes());
+        }
+        byte[] msg = new byte[32];
+        byte[] pubKey = new byte[32];
+        byte[] sig = new byte[64];
+
+        try {
```
aionick: All LGTM. Just wondering, is it worth it to return an error if the input length is larger than expected, since this may indicate that the user has incorrectly understood the format?

centrys-david: @aionick What do you think about creating a new `ResultCode` like `ExecutionResult.ResultCode.INVALID_INPUT`? Are these codes documented for the clients that use the precompiled contracts? So I know if I need to update the docs.

aionick: I'd like to look into this. From my understanding `ResultCode` is only used internally and never makes it out to the user. I may be wrong here. But I do think we need a little more flexibility in the precompileds for documenting errors. @yao-sun @AionJayT what do you guys think?

AionJayT: @centrys-david, we can create the case to handle the INVALID_INPUT. But it requires a few places change.
@aionick we definitely need to define how to use return code in case the developer misusing.

centrys-david: @AionJayT @aionick 

There are 2 cases here:

1. Smart contract client
I don't know how to throw/return an error from the precompiled contract that can be interpreted by the aion_fastvm and solidity compiler. The best practices are using `require` so there needs to be a documentation for the client to say something like:
```
...
require(hash.length == 32)
require(pub.lenght == 32)
...
edveirfy(hash, pub, sig1, sig2)
...
```

2. Java client 
If someone wants to integrate this precompiled contract on a lower level (eg. calling it directly) there is a minium amount of logging [here](https://github.com/aionnetwork/aion/pull/589/files/bcac62e015b7f595e713fbe04ed1d6a04d0198b7#diff-ae8e32d7dc609cad8c9618627777a278R57) that one can check on the kernel when developing. 

Still, I think documentation is our best choice for now and a new `ExecutionResult.ResultCode.INVALID_INPUT`

Let me know what you think.

modPrecompiled/src/org/aion/precompiled/ContractFactory.java
```diff
@@ -28,6 +28,7 @@
 import org.aion.mcf.core.AccountState;
 import org.aion.mcf.db.IBlockStoreBase;
 import org.aion.precompiled.contracts.ATB.TokenBridgeContract;
+import org.aion.precompiled.contracts.EDVerifyContract;
```
aion82: Ahmmed Aion Chowdhury 
From Bangladesh 
Aion Not only the name 
Aion is world famous Brand for Crypto world (Cryptocurrency)

modPrecompiled/src/org/aion/precompiled/ContractFactory.java
```diff
@@ -95,4 +100,11 @@ public static Address getTotalCurrencyContractAddress() {
         return Address.wrap(TOTAL_CURRENCY);
```
aion82: Faridgonj Bazar Chandpur chittagonj Bangladesh

AionJayT:
rebase to brach ed25519_verify

PR #598, opened 2018-08-15
modGui: verify Aion kernel is alive after launching it from GUI

aion-kelvin:
## Notice

## Description

- This PR is to fix an issue where if a kernel launched from the GUI immediately crashes/errors, the GUI enters a broken state (does not realize kernel is dead).   This is partly to address an issue raised by @aionick in PR #582 [https://github.com/aionnetwork/aion/pull/582] where a kernel launch from GUI hangs and can't be terminated by the GUI.  I was not able to reproduce the exact problem, but I suspect it is caused by the GUI entering a bad state when launching the kernel immediately fails.  A common reason for this is if you launch the kernel from the GUI, but there is already an instance of the kernel running, so the newly launched kernel immediately fails because it can't acquire a lock on the DB.  I also noticed that the kernel does not exit in this case [root cause: RuntimeException thrown in a static initializer hangs kernel -- https://stackoverflow.com/questions/7080715/exception-in-static-initializer-causes-program-to-hang].  In order to detect the "kernel dies upon launch" condition, I had to also fix that issue in this PR.

Fixes Issue # .

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- Existing tests pass (modified code touches those tests)
- Manual verification of new code

## Verification


aion-kelvin:
@AlexandraRoatis 

>I launched and terminated the kernel from the GUI under normal circumstances and got this error:

Indeed, it is harmless; it just looks a bit stupid to print out a bunch of exceptions when everything's fine.  I would try to fix it, but since we're close to release, I'm afraid of accidentally breaking something else, so probably safer to leave it how it is for now.  I'll keep it in mind for the future though.

modGui/src/org/aion/gui/model/TransactionProcessor.java
```diff
@@ -102,16 +103,24 @@ public TransactionProcessor(KernelConnection kernelConnection,
     }
 
     public Set<TransactionDTO> getLatestTransactions(final String address) {
-        backgroundExecutor.submit(this::processTransactionsFromOldestRegisteredSafeBlock);
+        try {
+            backgroundExecutor.submit(this::processTransactionsFromOldestRegisteredSafeBlock).get();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        } catch (ExecutionException e) {
+            e.printStackTrace();
```
AlexandraRoatis: it would be preferable to log these exceptions instead of printing them

PR #600, opened 2018-08-16
Separate test precompiled contract

aionick:
## Description

- Removes the precompiled contract for testing from `modPrecompiled` and places a new mocked contract factory in the fastVM that was using that test PC.
- See fastVM side of PR <a href="https://github.com/aionnetwork/aion_fastvm/pull/22">here</a>

Fixes Issue # N/A.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [x] Unit test.

## Testing

- our test suite

## Verification

### DIFF
--- a/modPrecompiled/src/org/aion/precompiled/ContractFactory.java
+++ b/modPrecompiled/src/org/aion/precompiled/ContractFactory.java
@@ -24,27 +24,24 @@ package org.aion.precompiled;
 
 import org.aion.base.db.IRepositoryCache;
 import org.aion.base.type.Address;
-import org.aion.base.type.IExecutionResult;
 import org.aion.base.vm.IDataWord;
 import org.aion.mcf.core.AccountState;
 import org.aion.mcf.db.IBlockStoreBase;
 import org.aion.precompiled.contracts.ATB.TokenBridgeContract;
-import org.aion.vm.AbstractExecutionResult.ResultCode;
 import org.aion.vm.ExecutionContext;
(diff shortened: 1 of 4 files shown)


modPrecompiled/src/org/aion/precompiled/ContractFactory.java
```diff
@@ -56,7 +53,8 @@ public ContractFactory(){}
      * @param track The repo.
      * @return the specified pre-compiled address.
      */
-    public static IPrecompiledContract getPrecompiledContract(ExecutionContext context,
+    @Override
+    public IPrecompiledContract getPrecompiledContract(ExecutionContext context,
         IRepositoryCache<AccountState, IDataWord, IBlockStoreBase <?, ?>> track) {
 
         switch (context.address().toString()) {
```
yulongaion: The `context.address()` will be the contract address, rather than destination address, in case of `DELEGATECALL` and `CALLCODE`.

qoire: Would that be intended behaviour? Unless we want to implement library precompiled contracts in the future...

yulongaion: Good point. If we're not going to implement precompiled library in the future, then should be fine; otherwise, a hardfork would be required to enable it.

qoire: I'm going to say this is intended behaviour for now.

Issue #543, opened 2018-07-06
Checksummed Addresses


qoire:
Posting here because not sure where else to post, this would affect moreso our frontend utilities. But here's the proposed code:

https://github.com/aionnetwork/aion/commit/46b4c1fdcbda765db90349065d4f02be06793d2b

The idea is to go with the original Eth implementation, using each bit as a determination as to whether an alphabetical character should be capitalized or not. This can keep our hashes short (if we want we can event reduce this down to a MD5.

qoire:
Another question is whether we want to include the ``0xA0`` header character in the checksum. I'm of the opinion that we won't be differentiating between ``0xa0`` and ``0xA0`` anyways, so its safe to include.

qoire:
As for where this impacts our kernel, _it would be nice_ if our keystore could start generating files that had the checksummed addresses instead of regular ones, and _if_ we were to detect a checksummed address through (for example) a JSON-RPC call, we would verify the validity.

tcrowe:
@qoire Thanks for your reply. I've ported this one to work with AION addresses in the js client: https://github.com/ethereum/web3.js/blob/1.0/packages/web3-utils/src/index.js#L240

It will be in the aion_web3 PR in a few days.

It would help if you generate me 6 addreses and checksum addresses from the java client. I will include that in my test suite.

I got these addressess from the Conquest test net then I generated checksum addresses. To my best understanding it's correct. Hopefully we can work together to prove it.

```
0xa07c95cc8729a0503c5ad50eb37ec8a27cd22d65de3bb225982ec55201366920
0xA07C95cC8729a0503C5ad50eb37eC8a27cD22D65dE3BB225982Ec55201366920

0xa05a3889b106e75baa621b8cc719679a3dbdd799afac1ca6b42d03dc93a23687
0xa05A3889b106e75baA621b8cC719679a3dBdD799afac1ca6B42D03DC93A23687

0xa0229e43f4a040f9fa6b2ab2f2cccc066025117def3414e08edbe7aee8e61647
0xA0229E43f4a040F9fa6B2AB2F2Cccc066025117def3414E08EdbE7aeE8e61647

0xa0dd16394f16ea21c8b45c00b2e43850ae7e8f00fe54789ddd1881d33b21df0c
0xa0Dd16394f16Ea21C8B45C00b2E43850aE7E8f00fe54789ddD1881D33B21DF0C

[...]

PR #609, opened 2018-08-22
Checksum

qoire:
## Notice

## Description

Integrates initial checksum code, this will serve as the reference for other implementations (#543). Currently not _used_ in the codebase itself.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [ ] Unit test.
- [x] Requires documentation update.

## Testing

Tested to match the our web3.js 1.0 implementation.

## Verification


qoire:
Cool, will do

aion-kelvin:
+1 to what @AionJayT  said

modCrypto/test/org/aion/crypto/ChecksumTest.java
```diff
@@ -0,0 +1,13 @@
+package org.aion.crypto;
+
+import org.aion.base.type.Address;
+import org.aion.base.util.ByteUtil;
```
aionick: these two imports can be removed

modCrypto/src/org/aion/crypto/AddressSpecs.java
```diff
@@ -25,4 +26,26 @@ private AddressSpecs() {}
         buf.put(HashUtil.h256(publicKey), 1, 31);
         return buf.array();
     }
+
+    public static String checksummedAddress(String address) {
+        assert address != null;
+        address = address.replace("0x", "");
```
aionick: If we are given a string with "0x" sitting in the middle of it, or sitting in multiple places, this replacement could strip the string back down to a legitimate address - do we want this kind of behaviour?

arajasek: I think this is an interesting question, though I don't have a clear idea of the gravity of the implications.

I assume it's easy enough to "fix", simply only replace starting "0x"s?

modCrypto/src/org/aion/crypto/AddressSpecs.java
```diff
@@ -25,4 +26,31 @@ private AddressSpecs() {}
         buf.put(HashUtil.h256(publicKey), 1, 31);
         return buf.array();
     }
+
+    public static String checksummedAddress(String address) {
```
aion-kelvin: Can we change the method signature to return Optional<String>?  And in places where we currently return null, return `Optional.empty()` instead?

modCrypto/src/org/aion/crypto/AddressSpecs.java
```diff
@@ -25,4 +26,31 @@ private AddressSpecs() {}
...
+        assert address.length() == 64;
+        byte[] h;
+        try {
+            h = HashUtil.h256(ByteUtil.hexStringToBytes(address));
+        } catch (Exception e) {
+            e.printStackTrace();
+            return null;
+        }
+        if (h == null) return null; //address is invalid
+        System.out.println(ByteUtil.toHexString(h));
```
aion-kelvin: println not needed

modCrypto/src/org/aion/crypto/AddressSpecs.java
aion-kelvin: I think the "0x not at the start of string" problem still needs to be fixed (also can yuo add a unit test to make sure this doesn't happen?)

Suggestion: `address.replaceFirst("^0x", "")`

https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replaceFirst(java.lang.String,%20java.lang.String)

modCrypto/src/org/aion/crypto/AddressSpecs.java
```diff
@@ -25,4 +27,31 @@ private AddressSpecs() {}
...
+        for (int i = 0; i < 64; i++) {
+            if (Character.isDigit(caddr[i]))
+                continue;
+
+            if (Character.isAlphabetic(caddr[i])) {
+                caddr[i] = b.get(i) ? Character.toUpperCase(caddr[i]) : Character.toLowerCase(caddr[i]);
+                continue;
+            }
+        }
+        return Optional.of(String.valueOf(caddr));
```
AlexandraRoatis: I think it would make sense to append the `0x` to the resulting address.

PR #625, opened 2018-09-05
Decouple PoW implementations from general components

centrys-paul:
## Description

Further refactor existing Aion code in order to decouple PoW implementations from reusable components. An interface was extracted from AionHub and used wherever possible instead of the implementation. 

We also defined the interfaces for the proposed consensus engine architecture based on strategies as defined in [this diagram](https://drive.google.com/file/d/1vk0Fi7yU8YoKvNhbHzvU3xT3oOO3_3sV/view).

The Java Api and Web3 Api were refactored to take as parameter in the constructor an `IGenericAionChain` so they can be instantiated with different implementations.

Transaction creation was refactored by delegating the responsibility of tx instantiation to a factory. By this, the API is now implementation-agnostic and can create any type of transactions according to the blockchain implementation it uses. 

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

The existing unit tests all pass, and we also performed end-to-end tests containing the transaction flow.
-

## Verification


centrys-paul:
Thanks for the review. I updated the PR with a few changes to check types before casting

PR #632, opened 2018-09-10
RPC API Performance Features for Bridge

ali-sharif:
## Description

Critical performance issues resolved for bridge:
* Parallel processing of RPC batch calls
* Implemented faster receipt retrieval options in the `ops_` API
* Fixed formatting of dashboard API call in PB API (`ApiAion0`)
* Fixed bug in `getBlockByHash()` where a non-existant block hash was throwing an exception (**Spec-Defined Behaviour**: return `null`)

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

* Javascript based testbench to verify all changes: https://github.com/ali-sharif/aion/blob/master/parallel-batch-tests_18.09.09.tar.bz

## Verification

### DIFF
--- a/modApiServer/src/org/aion/api/server/pb/ApiAion0.java
+++ b/modApiServer/src/org/aion/api/server/pb/ApiAion0.java
@@ -2072,7 +2072,9 @@ public class ApiAion0 extends ApiAion implements IApiAion {
 
                     // blocks requested in the future. return empty result
                     if (blkStart > latestBlkNum) {
-                        Message.rsp_getBlockDetailsByRange rsp = Message.rsp_getBlockDetailsByRange.newBuilder()
+                        Message.rsp_getBlockDetailsByRange rsp =
+                            Message.rsp_getBlockDetailsByRange
+                                .newBuilder()
                                 .addAllBlkDetails(new ArrayList<>())
                                 .build();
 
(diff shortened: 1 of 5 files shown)


modApiServer/src/org/aion/api/server/rpc/RpcProcessor.java
```diff
@@ -9,16 +9,22 @@
...
-    RpcMethods apiHolder;
+    private RpcMethods apiHolder;
+
+    private ExecutorService executor;
+    private CompletionService<JSONObject> batchCallCompletionService;
+    private final int SHUTDOWN_WAIT_SECONDS = 5;
 
     public RpcProcessor(List<String> enabled) {
         this.apiHolder = new RpcMethods(enabled);
+        executor = Executors.newFixedThreadPool(Math.min(Runtime.getRuntime().availableProcessors() * 2, 4));
```
AionJayT: @ali-sharif 
feel like we open ThreadPool in the different place within rpc server and the rpc processor
Can we manage these thread in the same place?

modApiServer/src/org/aion/api/server/pb/ApiAion0.java
```diff
@@ -2072,9 +2072,7 @@ public ApiAion0(IAionChain ac) {
 
                     // blocks requested in the future. return empty result
                     if (blkStart > latestBlkNum) {
-                        Message.rsp_getBlockDetailsByRange rsp =
-                            Message.rsp_getBlockDetailsByRange
-                                .newBuilder()
+                        Message.rsp_getBlockDetailsByRange rsp = Message.rsp_getBlockDetailsByRange.newBuilder()
```
AionJayT: do you use google java formatter to format the code? the ApiAion0 looks no code has been changed. can you rever the change if this PR has no related with ApiAion0?

modApiServer/src/org/aion/api/server/rpc/RpcProcessor.java
ali-sharif: Those two thread pools serve a different purpose. 

The ***existing*** thread pool in the RPC server is HTTP server implementation-specific (ie. undertow manages it's own thread pool, we have to manage Nano's thread pool, etc.). 

This ***new*** thread pool deals with parallelizing the processing of the RPC batch calls and I think this is the most appropriate place to allocate this thread pool, since this class is delegated the responsibility of handling the RPC call String payload by the HTTP server.

Let me know if there is another place in the class hierarchy where you think might be a more appropriate place for this thread pool to be instantiated.

modApiServer/src/org/aion/api/server/pb/ApiAion0.java
ali-sharif: The Google formatting changes made that function in ApiAion0 pretty unreadable so these formatting changes were just done so I could read that implementation. Will revert these changes now since they are un-related to this PR.

ali-sharif: Ok, removed all changes to file ApiAion0

PR #635, opened 2018-09-13
Sync enhancement: first version of Lightning sync

AlexandraRoatis:
## Description

The first version of Lightning sync that improves the block syncing functionality by requesting blocks out of order ahead of import time and storing them temporarily to disk until they can be imported. 

This implementation leverages the fact that under normal circumstances there is a lot of repeated work being done due to requesting the same blocks multiple times from different peers.

Experimental results show that the full sync time is reduced by a factor of up to 10x when the kernel has a large number of peers. It achieves around 2x to 5x speedup with 10 to 40 nodes.

Fixes Issue #27.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

- Existing suite of tests, plus unit tests for the critical functionality introduced by this implementation.

## Verification


AlexandraRoatis:
Note: one of the unit tests requires persistence which is why the LevelDB jar is added to the build file. I took note of this issue #636 and plan for an implementation of a persistent mock database in the future.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+     *
+     * @param props the database properties to be used in initializing the underlying databases
+     */
+    private void init(Properties props) {
+        // initialize status
+        this.status = new HashMap<>();
+
+        // create the level source
+        props.setProperty(Props.DB_NAME, LEVEL_DB_NAME);
+        this.levelDatabase = connectAndOpen(props, LOG);
```
AionJayT: the connectAndOpen has a chance to return null object. I think the error needs to be handled.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+        this.levelSource = new ObjectDataSource<>(levelDatabase, HASH_LIST_RLP_SERIALIZER);
+
+        // create the queue source
+        props.setProperty(Props.DB_NAME, QUEUE_DB_NAME);
+        this.queueDatabase = connectAndOpen(props, LOG);
```
AionJayT: same here.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+        this.queueSource = new ObjectDataSource<>(queueDatabase, BLOCK_LIST_RLP_SERIALIZER);
+
+        // create the index source
+        props.setProperty(Props.DB_NAME, INDEX_DB_NAME);
+        this.indexSource = connectAndOpen(props, LOG);
```
AionJayT: same here

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+        return status != null
+                && levelSource.isOpen()
+                && queueSource.isOpen()
+                && indexSource.isOpen();
+    }
+
+    private static final Serializer<List<byte[]>, byte[]> HASH_LIST_RLP_SERIALIZER =
+            new Serializer<>() {
+                @Override
+                public byte[] serialize(List<byte[]> object) {
```
AionJayT: I think it's good the including the jsr305 library to check the nonnull status (adding @Nonnull in front of the inputting args )

AionJayT: for the Serializer class, do we need to confirm the input data has no null object?

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+                if (existingQueueHash.isPresent()) {
+                    // using parent queue hash
+                    currentQueueHash = existingQueueHash.get();
+
+                    // append block to queue
+                    currentQueue = queueSource.get(currentQueueHash);
+                } // do not add else here!
+
+                // when no queue exists OR problem with existing queue
+                if (currentQueue == null || currentQueue.size() == 0) {
```
AionJayT: use "isEmpty()" instead of size() == 0

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+                queueSource.put(currentQueueHash, currentQueue);
+
+                // update status tracking
+                ByteArrayWrapper hash = ByteArrayWrapper.wrap(currentQueueHash);
+                QueueInfo info = status.get(hash);
+                if (info == null) {
+                    if (Arrays.equals(currentQueueHash, block.getHash())) {
+                        info =
+                                new QueueInfo(
+                                        currentQueueHash, block.getNumber(), block.getNumber());
```
AionJayT: why the constructor has two same args input?

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+    /**
+     * Stores a range of blocks in the pending block store for importing later when the chain
+     * reaches the needed height and the parent blocks gets imported. The functionality expects
+     * consecutive blocks, but ensures correct behavior even when the blocks are not consecutive.
+     */
+    public int addBlockRange(List<AionBlock> blocks) {
+        List<AionBlock> blockRange = new ArrayList<>(blocks);
+
+        // nothing to do when 0 blocks given
+        if (blockRange.size() == 0) {
```
AionJayT: size() == 0 -> isEmpty()

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AionJayT: https://stackoverflow.com/questions/1508975/why-is-list-size0-slower-than-list-isempty-in-java

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+            lock.writeLock().unlock();
+        }
+    }
+
+    /** Stores a block ranges with the first element determining the queue placement. */
+    private int addBlockRange(AionBlock first, List<AionBlock> blockRange) {
+
+        // skip if already stored
+        while (indexSource.get(first.getHash()).isPresent()) {
+            if (blockRange.size() == 0) {
```
AionJayT: size() == 0 -> isEmpty()

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+
+        // add element to queue
+        currentQueue.add(first);
+
+        // keep track of parent to ensure correct range
+        AionBlock parent = first;
+
+        AionBlock current;
+        // process rest of block range
+        while (blockRange.size() > 0) {
```
AionJayT: size() > 0 -> !isEmpty()

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+        // done with queue
+        queueSource.putToBatch(currentQueueHash, currentQueue);
+
+        // the number of blocks added
+        return stored;
+    }
+
+    /**
+     * @return the number of elements stored in the index database.
+     * @implNote This method is package private because it is meant to be used for testing.
```
AionJayT: how about put @VisibleForTesting in front of this method?

modP2p/src/org/aion/p2p/P2pConstant.java
```diff
@@ -48,5 +48,20 @@
     // write queue timeout
     WRITE_MSG_TIMEOUT = 5000,
 
-    BACKWARD_SYNC_STEP = 128;
+    REQUEST_SIZE = 24,
+
+    /** @implNote must be strictly larger than {@link #REQUEST_SIZE} */
+    LARGE_REQUEST_SIZE = 25,
```
AionJayT: can we just define REQUEST_SIZE + 1?

modAionImpl/src/org/aion/zero/impl/sync/TaskGetHeaders.java
```diff
@@ -70,22 +75,31 @@
         this.log = log;
     }
 
+    /** Checks that the peer's total difficulty is higher than the local chain. */
+    private boolean isHigherTotalDifficulty(INode n) {
+        return n.getTotalDifficulty() != null && n.getTotalDifficulty().compareTo(this.selfTd) >= 0;
```
AionJayT: is the ">=" an expecting behavior? cause the method name is higher than TD

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -64,204 +80,622 @@
...
     @Override
     public void run() {
         Thread.currentThread().setPriority(Thread.MAX_PRIORITY);
         while (start.get()) {
-
             BlocksWrapper bw;
             try {
                 bw = downloadedBlocks.take();
             } catch (InterruptedException ex) {
                 return;
```
AionJayT: Add error log in this exception.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -64,204 +80,622 @@
...
-                displayId,
-                (state != null ? state.getMode() : PeerState.Mode.NORMAL),
-                b.getShortHash(),
-                b.getNumber(),
-                b.getTransactionsList().size(),
-                importResult,
-                t2 - t1);
+        if (log.isDebugEnabled()) {
+            // printing sync mode only when debug is enabled
+            log.info(
```
AionJayT: since it's a debugging log. should put it as the debug level log.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -64,204 +80,622 @@
...
+                            "Loaded {} blocks from disk from level {} queue {} before filtering.",
+                            batchFromDisk.size(),
+                            entry.getKey(),
+                            level);
+                }
+
+                // filter already imported blocks
+                batchFromDisk = filterBatch(batchFromDisk, chain, importedBlockHashes);
+
+                if (batchFromDisk.size() > 0) {
```
AionJayT: use !isEmpty()

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -64,204 +80,622 @@
...
+                for (AionBlock b : batchFromDisk) {
+                    try {
+                        importResult = importBlock(b, "STORAGE", state);
+
+                        if (importResult.isStored()) {
+                            importedBlockHashes.put(ByteArrayWrapper.wrap(b.getHash()), true);
+
+                            batch++;
+
+                            if (last <= b.getNumber()) {
```
AionJayT: has any change to import more blocks when last < b.getNumber()?

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
```diff
@@ -74,23 +94,40 @@ public void run() {
...
+                            + this.networkStatus.getTargetBestBlockNumber() //
+                            + " b-hash="
+                            + Hex.toHexString(this.chain.getBestBlockHash()) //
+                            + "/"
+                            + this.networkStatus.getTargetBestBlockHash()
+                            + "";
 
             p2pLOG.info(status);
 
+            if (p2pLOG.isDebugEnabled()) {
```
AionJayT: Should it be a TRACE level? Cause this information can be huge.

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
```diff
@@ -111,4 +148,73 @@ public void run() {
...
+    public String dumpPeerStateInfo(Collection<INode> filtered) {
+        List<NodeState> sorted = new ArrayList<>();
+        for (INode n : filtered) {
+            PeerState s = peerStates.get(n.getIdHash());
+            if (s != null) {
+                sorted.add(new NodeState(n, s));
+            }
+        }
+
+        if (sorted.size() > 0) {
```
AionJayT: !isEmpty()

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+            List<byte[]> levelData = levelSource.get(levelKey);
+
+            if (levelData == null) {
+                LOG.error(
+                        "Corrupt data in PendingBlockStorage. Level (expected to exist) was not found.");
+                // level already missing so nothing to do here
+            } else {
+                List<byte[]> updatedLevelData = new ArrayList<>();
+
+                levelData.forEach(
```
aion-kelvin: I don't have strong feelings about it, but I've read that Collection.forEach(<lambda>) is generally slower than doing a plain old for(Object x : X) loop (unless you use `parallelStream()`).  Just wanted to point that out if performance is a big consideration.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+    }
+
+    /**
+     * Generates a number greater or equal to the given {@code current} number representing the base
+     * value for a subsequent LIGHTNING request. The returned base is generated taking into
+     * consideration the status updates from {@link #addStatusBlock(AionBlock)} and the {@code
+     * knownBest} value for the peer for which this functionality is requested.
+     *
+     * <p>The bases are generated in an optimistic continuous manner based on the following
+     * assumptions:
```
aion-kelvin: </p>

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+    private String statusToString() {
+        StringBuilder sb = new StringBuilder("Current status queues:\n");
+        for (QueueInfo i : status.values()) {
+            sb.append(i.toString());
+            sb.append('\n');
+        }
+        return sb.toString();
+    }
+
+    static class QueueInfo {
```
aion-kelvin: I think this can be private

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -64,204 +80,622 @@
...
+     * Utility method that takes a list of blocks and filters out the ones that are restricted for
+     * import due to pruning and the ones that have already been imported recently.
+     *
+     * @param blocks the list of blocks to be filtered
+     * @param chain the blockchain where the blocks will be imported which may impose pruning
+     *     restrictions
+     * @param imported the collection of recently imported blocks
+     * @return the list of blocks that pass the filter conditions.
+     */
+    static List<AionBlock> filterBatch(
```
aion-kelvin: `@VisibleForTesting`

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -1181,6 +1190,31 @@ public synchronized void storeBlock(AionBlock block, List<AionTxReceipt> receipt
         setBestBlock(block);
     }
 
+    @Override
+    public boolean storePendingStatusBlock(AionBlock block) {
```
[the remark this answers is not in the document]
AlexandraRoatis: JavaDoc added in d51f090.

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -1181,6 +1190,31 @@ public synchronized void storeBlock(AionBlock block, List<AionTxReceipt> receipt
...
+        return repository.getPendingBlockStore().addStatusBlock(block);
+    }
+
+    @Override
+    public int storePendingBlockRange(List<AionBlock> blocks) {
```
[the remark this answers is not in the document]
AlexandraRoatis: JavaDoc added in d51f090.

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -677,6 +686,16 @@ public void close() {
...
             }
 
+            try {
+                if (pendingStore != null) {
+                    pendingStore.close();
+                    LOGGEN.info("Pending block store closed.");
+                    pendingStore = null;
+                }
+            } catch (Exception e) {
+                LOGGEN.error("Exception occurred while closing the pending block store.", e);
```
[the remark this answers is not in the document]
AlexandraRoatis: The close method is called only at shutdown, so the kernel will not continue running. Data may get corrupted if the databases are not closed correctly, but this is not a big issue in the case of the pending store since any blocks that cannot be loaded from storage can be requested by the sync protocol.

modP2p/src/org/aion/p2p/P2pConstant.java
AlexandraRoatis: My intention is for this value to be significantly larger. Tests with 60 looked promising, but more experiments are needed to determine the potential drawbacks of such a large value and what would be an optimal constant. For now, I updated this value to 40 (1c6bf7b) and opened an issue for further experimentation (#648).

modAionImpl/src/org/aion/zero/impl/sync/TaskGetHeaders.java
AlexandraRoatis: Technically, since there is a time delay between receiving the peer status and performing this check, a peer with a total difficulty equal to the local one has likely already advanced to a higher TD.
Still, I will rename the method to make it less ambiguous.

modAionImpl/src/org/aion/zero/impl/sync/TaskGetHeaders.java
```diff
@@ -98,12 +112,38 @@ public void run() {
...
+                    // request far forward blocks
+                    if (state.getBase() > selfNumber + LARGE_REQUEST_SIZE
+                            // there have not been STEP_COUNT sequential requests
+                            && !state.isOverRepeatThreshold()) {
+                        size = LARGE_REQUEST_SIZE;
+                        from = state.getBase();
+                        break;
+                    } else {
+                        // transition to ramp down strategy
+                        state.setMode(THUNDER);
```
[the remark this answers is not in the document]
AlexandraRoatis: I've struggled with that problem for a good number of days. My previous attempts have resulted in a loss of performance for the overall sync process. I have a plan of attempting to add a peer state manager in a future PR.

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
AlexandraRoatis: It's similar to the p2p printout which is placed in debug. I don't think it belongs in trace logging, because it's infrequent. But if you have a strong preference for moving it to trace, I will make the change.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -64,204 +80,622 @@
...
-                    default:
-                        break;
+            } catch (Throwable e) {
+                log.error("<import-block throw> {}", e.toString());
+                if (e.getMessage() != null && e.getMessage().contains("No space left on device")) {
+                    log.error("Shutdown due to lack of disk space.");
+                    System.exit(0);
                 }
+                // TODO test and determine consequences: continue or break?
+                continue;
```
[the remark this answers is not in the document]
AlexandraRoatis: I agree, however, this means that I have to extract the functionality for checking the first block from the loop.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
AlexandraRoatis: Please check if you agree with my change. I am logging the error only when shutdown was not called since it is to be expected that this will get interrupted at shutdown.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+ * received status blocks and proposing (mostly non-overlapping) base values for the requests.
+ *
+ * @author Alexandra Roatis
+ */
+public class PendingBlockStore implements Flushable, Closeable {
+
+    private static final Logger LOG = AionLoggerFactory.getLogger(LogEnum.DB.name());
+    private static final Logger LOG_SYNC = AionLoggerFactory.getLogger(LogEnum.SYNC.name());
+
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
```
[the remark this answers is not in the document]
AlexandraRoatis: Ups! Forgot to add the lock on the read functionality. Fixed in 76868b5.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: The `<p>` is added by the google formatter. It does not require a closing tag: https://www.w3.org/TR/html401/intro/sgmltut.html#h-3.2.1

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+                    for (byte[] b : object) {
+                        infoList[i] = RLP.encodeElement(b);
+                        i++;
+                    }
+                    return RLP.encodeList(infoList);
+                }
+
+                @Override
+                public List<byte[]> deserialize(byte[] stream) {
+                    RLPList list = (RLPList) RLP.decode2(stream).get(0);
```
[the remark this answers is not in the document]
AlexandraRoatis: That is how `decode2` works. It returns a list where the first element contains the decoding.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: `RLP.encodeElement` and `RLP.encodeList` will produce an encoding even if the input data is null. Also, both serializers are private to the class which makes sure to give them correct data, so I would prefer not to add the overhead of the annotations.

AlexandraRoatis:
@aion-kelvin Regarding your general question:
1. the peer states are internal data, they contain some information from the peer, but cannot be changed by it;
2. there are some safeguards in place regarding abnormally high total difficulty and receiving blocks that are incorrect or were not requested;

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+    }
+
+    /**
+     * Initializes and opens the databases where the pending blocks will be stored.
+     *
+     * @param props the database properties to be used in initializing the underlying databases
+     */
+    private void init(Properties props) {
+        // initialize status
+        this.status = new HashMap<>();
```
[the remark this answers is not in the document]
AlexandraRoatis: This was an issue because I forgot to add the read locks. It should not happen with the code updates.

AlexandraRoatis:
Rebased on `master-pre-merge` to be able to add f3105f6.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AionJayT: I see

AlexandraRoatis:
Sorry! The first commit after the reviews is 8edc6f6.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: solved in 27b13a7

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: solved in 27b13a7

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: solved in 27b13a7

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: updated in 0742fff

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: updated in 0742fff

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: updated in 0742fff

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: updated in 0742fff

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
AlexandraRoatis: updated in 0742fff

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: The two parameters had different meanings, the first and last number in the queue. Also, I couldn't use two constructors with the same parameters. Refactored in 153fa7c.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: updated in 2bc2919

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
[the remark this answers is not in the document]
AlexandraRoatis: updated in b5bbd67

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
AlexandraRoatis: updated in b5bbd67

modAionImpl/src/org/aion/zero/impl/sync/TaskGetHeaders.java
AlexandraRoatis: updated in 22a2d69

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -64,204 +80,622 @@
...
+                            peerState.getBase());
                 }
 
-                switch (importResult) {
-                    case IMPORTED_BEST:
-                    case IMPORTED_NOT_BEST:
-                    case EXIST:
-                        {
-                            importedBlockHashes.put(ByteArrayWrapper.wrap(b.getHash()), true);
+                statis.update(getBestBlockNumber());
```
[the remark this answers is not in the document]
AlexandraRoatis: updated in 22a2d69

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -512,12 +512,21 @@ public boolean skipTryToConnect(long blockNumber) {
         return blockNumber > current + 32 || blockNumber < current - 32;
     }
 
+    /**
+     * If using TOP pruning we need to check the pruning restriction for the block.
+     * Otherwise, there is not prune restriction.
+     */
+    public boolean checkPruneRestriction() {
```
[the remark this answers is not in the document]
AlexandraRoatis: updated in 8edc6f6

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -62,6 +62,9 @@
 
     private TransactionStore<AionTransaction, AionTxReceipt, AionTxInfo> transactionStore;
 
+    // pending block store
+    protected PendingBlockStore pendingStore;
```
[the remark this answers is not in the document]
AlexandraRoatis: updated in 27b13a7

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+                    for (RLPElement aList : list) {
+                        res.add(new AionBlock(aList.getRLPData()));
+                    }
+                    return res;
+                }
+            };
+
+    /**
+     * Stores a single block in the pending block store for importing later when the chain reaches
+     * the needed height and the parent block gets imported. Is used by the functionality receiving
```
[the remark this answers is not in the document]
AlexandraRoatis: updated in 2bc2919

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+                // block already stored
+                return false;
+            }
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    /**
+     * @return the number of elements stored in the status map. * @implNote This method is package
```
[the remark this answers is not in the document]
AlexandraRoatis: updated in 2bc2919

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+        } else {
+            return status.get(ByteArrayWrapper.wrap(hash));
+        }
+    }
+
+    /**
+     * Stores a range of blocks in the pending block store for importing later when the chain
+     * reaches the needed height and the parent blocks gets imported. The functionality expects
+     * consecutive blocks, but ensures correct behavior even when the blocks are not consecutive.
+     */
```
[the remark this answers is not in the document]
AlexandraRoatis: updated in 2bc2919

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+            queueSource.flushBatch();
+
+            // the number of blocks added
+            return stored;
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    /** Stores a block ranges with the first element determining the queue placement. */
```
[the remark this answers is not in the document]
AlexandraRoatis: updated in 2bc2919

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+    private int addBlockRange(AionBlock first, List<AionBlock> blockRange) {
```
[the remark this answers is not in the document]
AlexandraRoatis: added a JavaDoc note in 2bc2919

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: updated in 2bc2919

modAionImpl/src/org/aion/zero/impl/sync/PeerState.java
```diff
@@ -83,39 +126,67 @@ public long getLastHeaderRequest() {
...
 
     public void resetLastHeaderRequest() {
+        this.state = State.INITIAL;
         this.lastHeaderRequest = 0;
     }
 
-    public int getRepeated() {
-        return repeated;
+    public boolean isOverRepeatThreshold() {
+        return repeated >= STEP_COUNT;
```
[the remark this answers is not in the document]
AlexandraRoatis: renamed in 2bc2919

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AlexandraRoatis: I added the `@VisibleForTesting` tag in 2bc2919 since one of the testing methods returns this type of object

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
AlexandraRoatis: you're right, `last` will be at most equal to the block number; correction made in b5bbd67

modAionImpl/src/org/aion/zero/impl/sync/TaskGetHeaders.java
```diff
@@ -98,12 +112,38 @@ public void run() {
...
+                    }
+                }
+            case THUNDER:
+                {
+                    // there have not been STEP_COUNT sequential requests
+                    if (!state.isOverRepeatThreshold()) {
+                        state.setBase(selfNumber);
+                        size = LARGE_REQUEST_SIZE;
+                        from = Math.max(1, selfNumber + 1 - 4);
```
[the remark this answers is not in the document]
AlexandraRoatis: I added constants (with JavaDoc) for this value in db998d2

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,816 @@
...
+                    for (AionBlock b : object) {
+                        infoList[i] = b.getEncoded();
+                        i++;
+                    }
+                    return RLP.encodeList(infoList);
+                }
+
+                @Override
+                public List<AionBlock> deserialize(byte[] stream) {
+                    RLPList list = (RLPList) RLP.decode2(stream).get(0);
```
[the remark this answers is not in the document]
AlexandraRoatis: same as above

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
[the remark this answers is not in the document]
AlexandraRoatis: Incorrect serializations are unlikely to occur since the class controls both the serializer and deserializer and even null objects will be serialized correctly.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -0,0 +1,888 @@
...
+            } finally {
+                try {
+                    indexSource.close();
+                } catch (Exception e) {
+                    LOG.error("Not able to close the pending blocks index database:", e);
+                } finally {
+                    lock.writeLock().unlock();
+                }
+            }
+        }
```
AionJayT: can we rewrite this code section with one try catch?
can use a string object to represent the status in case of exception happens.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -56,212 +73,638 @@
...
-                            // since last import worked skipping the batch
-                            batch.clear();
-                            log.info("Forward skip.");
-                            break;
-                        }
-                    default:
-                        break;
+    /** @implNote This method is called only when state is not null. */
+    private PeerState processBatch(PeerState givenState, List<AionBlock> batch, String displayId) {
+        // make a copy of the original state
```
AionJayT: How about add require check ?
Objects.requireNonNull(givenState)

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
AionJayT: then we do 3 separate try catch and put the unlock at the end of the method.

AlexandraRoatis:
Full sync duration experiments performed on two identical machines (512 GB SSD, 4 Core, 16GB RAM) using the `mastery` network:

  | Version | Average active peers | Sync Duration | Start Time | Arrived at Head | Block Nb. At Head |  
-- | -- | -- | -- | -- | -- | -- | --
Release | dcaf9e8 | 34 | 08:46:58 | 2018-09-28 18:02:09 | 2018-09-29 02:49:07 | 572,239 | Machine 1
Release | dcaf9e8 | 33 | 08:48:23 | 2018-09-28 18:02:09 | 2018-09-29 02:50:33 | 572,251 | Machine 2
Lightning Sync V1 | d539cda | 32 | 01:22:30 | 2018-09-28 18:02:09 | 2018-09-28 19:24:39 | 569,616 | Machine 1
Lightning Sync V1 | d539cda | 31 | 01:15:04 | 2018-09-28 18:02:10 | 2018-09-28 19:17:13 | 569,570 | Machine 2

AlexandraRoatis:
Full sync duration experiments performed on two identical machines (512 GB SSD, 4 Core, 16GB RAM) using the `mastery` network:

  | Version | Average active peers | Sync Duration | Start Time | Arrived at Head | Block Nb. At Head |  
-- | -- | -- | -- | -- | -- | -- | --
Release | dcaf9e8 | 34 | 08:46:58 | 2018-09-28 18:02:09 | 2018-09-29 02:49:07 | 572,239 | Machine 1
Release | dcaf9e8 | 33 | 08:48:23 | 2018-09-28 18:02:09 | 2018-09-29 02:50:33 | 572,251 | Machine 2
Lightning Sync V1 | d539cda | 32 | 01:22:30 | 2018-09-28 18:02:09 | 2018-09-28 19:24:39 | 569,616 | Machine 1
Lightning Sync V1 | d539cda | 31 | 01:15:04 | 2018-09-28 18:02:10 | 2018-09-28 19:17:13 | 569,570 | Machine 2

modAionImpl/src/org/aion/zero/impl/sync/SyncStats.java
```diff
@@ -33,15 +33,15 @@
...
+final class SyncStats {
 
     private long start;
 
     private long startBlock;
 
     private double avgBlocksPerSec;
 
-    SyncStatics(long _startBlock){
+    SyncStats(long _startBlock){
```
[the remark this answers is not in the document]
AlexandraRoatis: Candidate for a bounty?

PR #641, opened 2018-09-14
Zmq secure connect implementation 

AionJayT:
## Notice

## Description

The implementation allows the zmq server setup the secure connection by the build-in Curve algo.
To enable the secure connection, the user can enable the settings in the config.xml "secure-connect".
Also, the user needs to generate the curve keypair ( by using ./aion.sh -zs). And give the public key to the API client side. The client API need to upgrade to the v0.1.13 or later version.

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

it has been tested by the wallet team.

## Verification


modApiServer/src/org/aion/api/server/zmq/ProtocolProcessor.java
```diff
@@ -144,6 +186,39 @@ public void run() {
         }
     }
 
+    private void loadCurveKeyPair() {
+        List<File> files = getFiles();
```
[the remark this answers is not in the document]
AlexandraRoatis: Moreover, since generating the keys does not require any user input, why not simply generate them for the user when the fils are missing?

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -786,7 +853,7 @@ private static boolean copyRecursively(File src, File target)
         }
         else {
             try {
-                Files.copy(src, target);
+                com.google.common.io.Files.copy(src, target);
```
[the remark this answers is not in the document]
AionJayT: there are 2 different Files classes using in this class

modApiServer/src/org/aion/api/server/zmq/ProtocolProcessor.java
```diff
@@ -53,24 +61,41 @@
...
-    private AtomicBoolean shutDown = new AtomicBoolean();
-
+    private static final String CURVEKEY_PATH;
+    private static final Path PATH;
     private static final long zmqHWM = 100_000;
     private static final int SOCKETID_LEN = 5;
     private static final int SOCKET_RECV_TIMEOUT = 3000;
 
+    static {
+        String storageDir = System.getProperty("local.storage.dir");
```
aion-kelvin: Can we do the following to avoid the reliance on the static block:

- make it so that `storageDir` is an argument to the constructor
- make `CURVEKEY_PATH` and `PATH` non-static
- move the logic of setting CURVEKEY_PATH and PATH into constructor
- add either another constructor or a static method, that will call the constructor, and pass in `System.getProperty('local.storage.dir') as the value for storageDir

modApiServer/src/org/aion/api/server/zmq/ProtocolProcessor.java
```diff
@@ -144,6 +186,39 @@ public void run() {
         }
     }
 
+    private void loadCurveKeyPair() {
+        List<File> files = getFiles();
+        String nextLoad = "";
+        for (File f : files) {
```
aion-kelvin: Is there a benefit to this fall-back/fuzzy logic?  Why not just have two hard-coded file names it looks for?

modMcf/src/org/aion/mcf/config/CfgApiZmq.java
```diff
@@ -38,19 +38,23 @@
...
 
+    public static final String ZMQ_KEY_DIR = "zmq_keystore";
+
     CfgApiZmq() {
         this.active = true;
         this.ip = "127.0.0.1";
         this.port = 8547;
         this.filtersEnabled = true;
         this.blockSummaryCacheEnabled = false;
+        this.secureConnectEnabled = false;
```
aion-kelvin: optional: might be safer to default it to true (after all, it's true by default in config.xml)

modApiServer/src/org/aion/api/server/zmq/ProtocolProcessor.java
AionJayT: I was thinking the user will generate several keypairs in case need to switch the keyset. The public key also needs to be updated to the client side, then the connection will be setup.

modMcf/src/org/aion/mcf/config/CfgApiZmq.java
AionJayT: the will notice in the document and wiki later.

Issue #642, opened 2018-09-18
Very high values when eth.estimateGas requested


gurdeepakjoshi:
**Which Aion version?**: 0.3.1.46f09bac
**Which operating system?**: Linux 
**Which distributor and version?**: Ubuntu 16.04 LTS 64 bit
**How installed?**: binaries
**Are you fully synchronized?**: yes
 **Did you try to restart the node?**: yes

Actual Behaviour
- Launch aion node
- cd web3/
- node console.js
- aion-127.0.0.1:8545> eth.estimateGas({data: "0x000"})
13914991129094744000
It returns a very high estimate which makes the contract deployment infeasible.

Expected Behaviour
estimateGas should return realistic estimate such as 2000000

FYI
If you run same node locally (private) without syncing with external nodes then this issue doesn't occur. This issue is only occuring on fully synced node.

qoire:
@AionJayT might be important for us to check that we can reproduce this.

From my side:

I was unable to reproduce this issue with a locally synced node (as suggested by Gurdeep). I also was informed by another colleague that it seemed to have to do with his keystore being placed in the keystore folder vs not. I tried this as well and was not able to reproduce the error.

qoire:
@gurdeepakjoshi Just to be clear, when you say fully synced, are you referring to ``mainnet`` or ``mastery``?

gurdeepakjoshi:
> @gurdeepakjoshi Just to be clear, when you say fully synced, are you referring to `mainnet` or `mastery`?

I meant mastery.

arajasek:
I was able to reproduce this on mastery using both v0.3.0.17d0f332, as well as the latest dev. I did not need to sync to completion in order to reproduce the bug.

I stepped through the code, and it looks like the wonky behaviour is happening inside the VM. @AionJayT is following up with the VM team.

yulongaion:
Quick fix would be updating the `AionTransaciton` constructor to not set the `from` variable.

Sine this issue won't affect consensus, put it at low-to-median priority.

AionJayT:
some of the contract calls do need to verify the sender, therefore, I think the better solution is putting an address when the call doesn't have the sender address or just return an error in the API layer.

yulongaion:
I saw the transaction is signed by a random address. If we don't set the `from` variable, the `getFrom()` method will decode the address from the signature.

arajasek:
PR #646 adds the Zero Address if no address is provided, resolves the issue.

PR #646, opened 2018-09-20
Put Zero address if no from address supplied

arajasek:
## Notice

## Description

If the API's estimateGas function is called without a from address, the VM uses the wrong offset when serialising. The fix is to simply use the Zero Address instead.

Fixes Issue #642 

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

All existing tests pass

## Verification

### DIFF
--- a/modApiServer/src/org/aion/api/server/ApiAion.java
+++ b/modApiServer/src/org/aion/api/server/ApiAion.java
@@ -462,8 +462,10 @@ public abstract class ApiAion extends Api {
     }
 
     protected long estimateNrg(ArgTxCall params) {
-        AionTransaction tx = new AionTransaction(params.getNonce().toByteArray(), params.getFrom(), params.getTo(),
+        Address fromAddr = (params.getFrom().equals(Address.EMPTY_ADDRESS())) ? Address.ZERO_ADDRESS() : params.getFrom();
+        AionTransaction tx = new AionTransaction(params.getNonce().toByteArray(), fromAddr, params.getTo(),
                 params.getValue().toByteArray(), params.getData(), params.getNrg(), params.getNrgPrice());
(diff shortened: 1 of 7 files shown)


modApiServer/src/org/aion/api/server/ApiAion.java
```diff
@@ -462,8 +462,10 @@ protected TxRecpt getTransactionReceipt(byte[] txHash) {
     }
 
     protected long estimateNrg(ArgTxCall params) {
-        AionTransaction tx = new AionTransaction(params.getNonce().toByteArray(), params.getFrom(), params.getTo(),
+        Address fromAddr = (params.getFrom().equals(Address.EMPTY_ADDRESS())) ? Address.ZERO_ADDRESS() : params.getFrom();
```
aion-kelvin: Maybe I'm reading this wrong, but isn't this just the same as `Address fromAddr = params.getFrom()` ?

Maybe the condition ternary is supposed to be params.getFrom.isEmpty() ?

arajasek: No, because the EMPTY_ADDRESS and ZERO_ADDRESS are distinct things, so it's not just the same as params.getFrom().

isEmpty() isn't a method in the Address class, as far as I can tell.

arajasek: done

arajasek:
Good for final review now! :)

AionJayT:
please make same PR for master-pre-marge

arajasek:
Created a new PR for master-pre-merge, this PR can be ignored and closed

Issue #627, opened 2018-09-06
unable to set absolute db path


aion-chris:
_Before filing a new issue, please **provide the following information**._

> I'm running:
>
> - **Which Aion version?**: 0.3.1.8dc68705f4c5c52008ea8a09074480fd9ce35fe3
> - **Which operating system?**: Linux
> - **Which distributor and version?**: Ubuntu 16.04
> - **How installed?**: via from source
> - **Are you fully synchronized?**: no
> - **Did you try to restart the node?**: yes

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._
unable to absolute db path

config.xml 
```
<path>/some_absolute_db_path</path>
```
src
modDbImpl/src/org/aion/db/impl/DatabaseFactory.java -> connectBasic -> info.getProperty(Props.DB_PATH)

output
```
18-09-06 17:35:38.123 DEBUG DB   [main]: db factory: /home/chris/aion/some_absolute_db_path
```

AlexandraRoatis:
This is not a bug, but a new feature request. Check the documented behavior: https://github.com/aionnetwork/aion/wiki/Database#database-location

You can create a symbolic link (`ln -s`) to the path you want.

qoire:
@kzeine candidate for bounty

AlexandraRoatis:
solved in #650

Issue #524, opened 2018-06-11
Add `--network` and `--datadir` options


yulongaion:
It would be nice to have the following options:
* `--network`: specify which network to connect, e.g. mainnet or testnet
* `--datadir`: specify the location of database dir

AlexandraRoatis:
Completed in #650. More details on behavior described in https://github.com/aionnetwork/aion/pull/650#discussion_r221070954

PR #650, opened 2018-09-25
CLI refactoring

AlexandraRoatis:
## Description

1. Using the [picocli](https://picocli.info/) library for parsing CLI input parameters.
2. Support for absolute path use for `log` and `database` folders #627.
3. Correct and consistent behavior for the `--network` and `--datadir` options #524.
4. Advanced `Network` enum with support for the `mastery` test net and custom networks.
5. Compatibility with old kernels when the above options are not used and the `config.xml` and `genesis.json` files are encountered at the old location.
6. Extensive unit tests for the new functionality.

## Type of change

- [x] Bug fix.
- [x] New feature.
- [x] Enhancement.
- [x] Unit test.
- [x] Requires documentation update.

## Testing

- Existing and newly added unit tests.
- The following manual commands:
```
# saving initial config (withut ids)
rsync -rv config/ iniconfig/

###### will exit the kernel

### help tests

./aion.sh -n testnet -d data -other
./aion.sh -other
./aion.sh -h

### version tests

./aion.sh -v
./aion.sh --version

### config tests

# compatible with old kernels
rsync -rv iniconfig/mainnet/ config/ #; grep "id" config/config.xml
./aion.sh -c # writes to config/config.xml
grep "id" config/config.xml

# for new kernels
rm -fr config/config.xml config/genesis.json
rm -fr mainnet

rsync -rv iniconfig/mainnet/ config/mainnet/  #; grep "id" config/mainnet/config.xml
./aion.sh -c # writes to config/mainnet/config.xml
grep "id" config/mainnet/config.xml

rsync -rv iniconfig/mainnet/ config/mainnet/ #; grep "id" config/mainnet/config.xml
./aion.sh -c mainnet # writes to config/mainnet/config.xml
grep "id" config/mainnet/config.xml

rsync -rv iniconfig/mainnet/ config/mainnet/ #; grep "id" config/mainnet/config.xml
./aion.sh -c incorrect # writes to config/mainnet/config.xml
grep "id" config/mainnet/config.xml

rm -fr mastery

rsync -rv iniconfig/mastery/ config/mastery/ #; grep "id" config/mastery/config.xml
./aion.sh -c mastery # writes to config/mastery/config.xml
grep "id" config/mastery/config.xml

rsync -rv iniconfig/mastery/ config/mastery/ #; grep "id" config/mastery/config.xml
./aion.sh -c testnet # writes to config/mastery/config.xml
grep "id" config/mastery/config.xml

mkdir mainnet ; mkdir mainnet/config ; rsync -rv iniconfig/mainnet/ mainnet/config/
./aion.sh -c # writes to mainnet/config/config.xml
grep "id" mainnet/config/config.xml
./aion.sh -c mainnet # writes to mainnet/config/config.xml
grep "id" mainnet/config/config.xml
./aion.sh -c incorrect # writes to mainnet/config/config.xml
grep "id" mainnet/config/config.xml

### info tests

# compatible with old kernels
rsync -rv iniconfig/mainnet/ config/
./aion.sh -i # reads config/config.xml

# for new kernels
rm -fr config/config.xml config/genesis.json
rm -fr mainnet

rsync -rv iniconfig/mainnet/ config/mainnet/
./aion.sh -i # reads config/mainnet/config.xml
mkdir mainnet ; mkdir mainnet/config ; rsync -rv iniconfig/mainnet/ mainnet/config/
./aion.sh -i # reads mainnet/config/config.xml

rm -fr mainnet
./aion.sh -i -n mainnet # reads config/mainnet/config.xml
mkdir mainnet ; mkdir mainnet/config ; rsync -rv iniconfig/mainnet/ mainnet/config/
./aion.sh -i -n mainnet # reads mainnet/config/config.xml

rm -fr data/mastery
./aion.sh -i -n testnet -d data # reads config/mastery/config.xml
mkdir data; mkdir data/mastery ; mkdir data/mastery/config ; rsync -rv iniconfig/mainnet/ data/mastery/config/
./aion.sh -i -n testnet -d data # reads data/mastery/config/config.xml

###### will run the kernel

rsync -rv iniconfig/mainnet/ config/
./aion.sh # reads from config/config.xml

rm -fr mainnet config; rsync -rv  iniconfig/ config/
./aion.sh # reads from config/mainnet/config.xml
# do not delete data for next test

./aion.sh # reads from mainnet/config/config.xml

rm -fr data config; rsync -rv  iniconfig/ config/
./aion.sh -n testnet -d data # reads from config/mastery/config.xml
# do not delete data for next test

./aion.sh -n testnet -d data # reads from data/mastery/config/config.xml

###### using data from last execution
./aion.sh -n testnet -d data --dump-blocks

./aion.sh -n testnet -d data -r 100

./aion.sh -n testnet -d data --state TOP

./aion.sh ac -n testnet -d data # store account in data/mastery/keystore

./aion.sh --dump-state -n testnet -d data
```

## Verification


modAionImpl/src/org/aion/zero/impl/cli/Arguments.java
```diff
@@ -0,0 +1,261 @@
...
+            description = "display information")
+    private boolean info = false;
+
+    @Option(
+            names = {"-v"},
+            description = "display version")
+    private boolean version = false;
+
+    @Option(
+            names = {"--version"},
```
aion-kelvin: Can we call this --versionTag?

modAionImpl/src/org/aion/zero/impl/cli/Arguments.java
```diff
@@ -0,0 +1,261 @@
...
+            description = "display version tag")
+    private boolean versionTag = false;
+
+    // create ssl certificate
+    @Option(
+            names = {"sc", "-s create"},
+            arity = "0..2",
```
aion-kelvin: I'm guessing there's no way to say arity is 0 or 2 (instead of 0, 1, 2) ?

modAionImpl/src/org/aion/zero/impl/cli/Arguments.java
```diff
@@ -0,0 +1,261 @@
...
+            names = {"--db-compact"},
+            description = "if using leveldb, it triggers its database compaction processes")
+    private boolean dbCompact;
+
+    /** Compacts the account options into specific commands. */
+    public static String[] preProcess(String[] arguments) {
+        List<String> list = new ArrayList<>();
+
+        int i = 0;
+        while (i < arguments.length) {
```
aion-kelvin: optional: for loop seemed more intuitive to me...

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -52,496 +57,511 @@
...
+                if (!setDirectory(options.getDirectory(), cfg)) {
+                    return ERROR;
+                }
+                // no return -> allow for other parameters combined with -d
+            }
 
-                        }
+            // reading from correct config file
+            File configFile = cfg.getExecConfigFile();
+            if (!configFile.exists()) {
```
aion-kelvin: if the given config file doesn't exist, can we either:

1) exit with an error
2) use the default but also print out a warning that the given input was not valid so we used the initial config (also, is it possible that the default config doesn't exist either?)

I feel like the user could get confused is he/she inputs something invalid and the program silently ignores it and does its own thing

modAionImpl/src/org/aion/zero/impl/config/Network.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param _name network name
+     * @param _identifier network identifier
+     */
+    Network(String _name, int _identifier) {
+        this.name = _name;
+        this.identifier = _identifier;
```
aion-kelvin: is this the same thing as chainId?  if so, I think that'd be better name

modAionImpl/src/org/aion/zero/impl/config/Network.java
```diff
@@ -0,0 +1,129 @@
...
+            return Network.MASTERY;
+        }
+
+        for (Network net : Network.values()) {
+            if (netStr.equals(net.toString())) {
+                return net;
+            }
+        }
+
+        return null;
```
aion-kelvin: would it make more sense to throw IllegalArgumentException?

modAionImpl/src/org/aion/zero/impl/config/Network.java
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Utility method that determines the correct network based on the given identifier.
+     *
+     * @param identifier a positive integer value representing the network identifier
+     * @return the network object corresponding to the int value or null when the value is not
+     *     mapped to an object.
+     */
+    public static Network determineNetwork(int identifier) {
+        if (identifier < 0) {
+            return null;
```
aion-kelvin: would IllegalArgumentException make more sense?

modAionImpl/src/org/aion/zero/impl/config/Network.java
```diff
@@ -0,0 +1,129 @@
...
+        return identifier;
+    }
+
+    /**
+     * Utility method that determines the correct network based on the given name.
+     *
+     * @param network a string value representing the network name
+     * @return the network object corresponding to the string value or null when the value is not
+     *     mapped to an object.
+     * @implNote There may be several test networks active. The term {@code testnet} is defined to
```
aion-kelvin: would it be more convenient to just call it mastery instead of testnet then?

I'm just thinking in the future, if there's another testnet that becomes the "main" testnet, then all the old docs/scripts that refer to mastery as testnet will become confusing

modAionImpl/src/org/aion/zero/impl/config/Network.java
```diff
@@ -0,0 +1,129 @@
...
+        return output.substring(1, output.length() - 1);
+    }
+
+    /**
+     * Generates a custom network with the given identifier.
+     *
+     * @param identifier a positive integer value representing the network identifier
+     * @return a custom network object with the given identifier.
+     */
+    public static Object getCustomNet(int identifier) {
```
aion-kelvin: why Object?

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
```diff
@@ -0,0 +1,1167 @@
...
+@RunWith(JUnitParamsRunner.class)
+public class CliTest {
+
+    private final Cli cli = new Cli();
+    private final CfgAion cfg = CfgAion.inst();
+    private final Cli mockCli = spy(cli);
+
+    // base paths
+    private static final String BASE_PATH = System.getProperty("user.dir");
+    private static final File MAIN_BASE_PATH = new File(BASE_PATH, "mainnet");
```
aion-kelvin: You could also use JUnit's TemporaryFolder to generate a temporary folder and set your base paths to that folder.  Then you won't need to manually delete stuff in your `@After `

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
```diff
@@ -0,0 +1,1167 @@
...
+    public void testImportNonPrivateKey(String option) {
+        String fakePKey = Hex.toHexString("random".getBytes());
+
+        assertThat(mockCli.call(new String[] {option, fakePKey}, cfg)).isEqualTo(ERROR);
+
+        assertThat(mockCli.call(new String[] {option, "hello"}, cfg)).isEqualTo(ERROR);
+    }
+
+    // Methods below taken from FileUtils class
+    private static boolean copyRecursively(File src, File target) {
```
aion-kelvin: We should probably look into using Apache Commons IO (or something similar) for these functions instead of just copy pasting them everywhere...

modBoot/src/org/aion/Aion.java
```diff
@@ -121,29 +110,31 @@ public static void main(String args[]) {
...
-        AionLoggerFactory
-            .init(cfg.getLog().getModules(), cfg.getLog().getLogFile(), cfg.getLog().getLogPath());
-        Logger genLog = AionLoggerFactory.getLogger(LogEnum.GEN.name());
-
         String[] filePath = new String[7];
         // Log/Database path
         if (!cfg.getLog().getLogFile()) {
-            System.out.println("Logger disabled; to enable please check log settings in config.xml");
+            System.out.println("Logger disabled; to enable please update log settings in config.xml and restart kernel.");
+            filePath[0] = "« disabled »";
```
aion-kelvin: »

modAionImpl/src/org/aion/zero/impl/cli/Arguments.java
AlexandraRoatis: Nope, there is not. Because of this reason, I had to edit the help printout as well.

modMcf/src/org/aion/mcf/config/Cfg.java
```diff
@@ -136,11 +129,242 @@ public void setConsensus(CfgConsensus _consensus) {
         this.consensus = _consensus;
     }
 
+    /* ------------ execution path management ------------ */
```
aion-kelvin: Supporting the old directory structure as well as the new one is definitely convenient and I'm sure users will appreciate it.  Is this something we should continue to support into the future though?

I think at some point (not in this PR) we should make a converter that converts the old structure to the new one so we don't have the burden of dealing with the two code paths.  The transition could go like this:

1. Make the script and release it
2. Add message in kernel so if it detects old structure, print a message saying they need to migrate it using the script at some point, because we'll stop supporting it 
3. Wait a while, deprecate the code that supports it

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
AlexandraRoatis: I will explain the reasoning behind that code below.
First, keep in mind the following 3 points:
1. there exists a **base directory** where all the code and the `aion.sh` reside;
2. we allow setting an **execution path** with the `--datadir` option and/or the `--network` option;
3. we must maintain compatibility with old kernel executables.
So, in order to make these features compatible, I've opted for the following:
a) If I find a `config.xml` and `genesis.json` at the old location, I will attempt to execute the CLI command or run the kernel as previous kernels would, in the **base directory**.
b) But, if the user calls the kernel with either the `--datadir value`, `--network value` or `--config value` option, then it will take precedence over old executable style and behave in the following way:
-> we create an **execution path** which will be either a directory in the base path with the network name (to ensure that if I run different networks from the same base path, they will not clash with eachother) or the directory given in the datadir option plus the network;
-> we copy the configuration files to the **execution path** to save the configuration;
-> we create the `keystore`, `log` and `database` directories, if they do not alreasy exist.

AlexandraRoatis: Because of this setup, there are in fact 3 places from where a kernel can read it's configuration:
1. `config/config.xml` like old kernels
2. `config/network/config.xml` for new kernels with new execution paths
3. `network/config/config.xml` for new kernels where the execution path already exists

AlexandraRoatis: I agree that this will need to be well documented when we introduce the `--datadir` and `--network` options.

modAionImpl/src/org/aion/zero/impl/config/Network.java
AlexandraRoatis: not in this case, because if this method returns `null` the calling method will add the default `mainnet` setting allowing for the kernel to run

modAionImpl/src/org/aion/zero/impl/config/Network.java
AlexandraRoatis: honestly, the question would be more along the lines of: Should we prevent negative chain ids for custom networks? I don't have strong feelings about it.

modAionImpl/src/org/aion/zero/impl/cli/Arguments.java
aion-kelvin: Oh, if there's pre-existing behaviour based on it, I guess we should keep it the unchanged then

modAionImpl/src/org/aion/zero/impl/config/Network.java
AlexandraRoatis: I disagree here. The alternative is to disallow the keyword which would end up with someone calling for `testnet` and getting the default `mainnet` without noticing the change. 

Additionally, an advantage to having this kind of default set is exactly for when a test network is being phased out. A user can still run on `mastery` by giving the keyword, but the default for new users will be the new network, enforcing the transition is a way. Updating documentation is simply a price to be paid when transitioning to a new test network.

modAionImpl/src/org/aion/zero/impl/config/Network.java
AlexandraRoatis: bug

modAionImpl/src/org/aion/zero/impl/cli/Arguments.java
AionJayT: Most of applications use `--version`  to check the version info.
Therefore I prefer don't change it.

modAionImpl/src/org/aion/zero/impl/config/Network.java
AionJayT: I think it does matter keep the chain id as a positive number. More easy to recognize and the format in the future spoke network.

AlexandraRoatis:
confirmed that the full list of commands in the PR description works correctly after the bug fixes in 47b2266

AlexandraRoatis:
Reviewers are encouraged to check the commands as well and perform additional tests, as they see fit ;)

AlexandraRoatis:
@aion-kelvin regarding the `./aion.sh al ac` issue we will document the behavior and bounty out the incompatibility checks

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
AlexandraRoatis: I created an issue for this refactoring #660

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
AlexandraRoatis: I created an issue for this refactoring #660

PR #663, opened 2018-10-05
Sync: Enhanced TaskImportBlocks

AlexandraRoatis:
## Description

* Improves the `TaskImportBlocks` functionality by delegating some storage operations to parallel threads.
* Optimized locks for the `PendingBlockStore`.
* Corrected the circumstances under which blocks are skipped by the import functionality.

Continued work on Issue #27.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing test suite and running the kernel

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
@@ -40,7 +40,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.aion.base.db.Flushable;
 import org.aion.base.db.IByteArrayKeyValueDatabase;
@@ -84,14 +86,15 @@ public class PendingBlockStore implements Flushable, Closeable {
     private static final Logger LOG = AionLoggerFactory.getLogger(LogEnum.DB.name());
     private static final Logger LOG_SYNC = AionLoggerFactory.getLogger(LogEnum.SYNC.name());
 
-    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+    private final ReadWriteLock databaseLock = new ReentrantReadWriteLock();
+    private final Lock internalLock = new ReentrantLock();
 
     // database names
     private static final String LEVEL_DB_NAME = "level";
     private static final String QUEUE_DB_NAME = "queue";
     private static final String INDEX_DB_NAME = "index";
 
-    // data sources
+    // data sources: with access managed by the `databaseLock`
     /**
      * Used to map a level (blockchain height) to the queue identifiers that start with blocks at
      * that height.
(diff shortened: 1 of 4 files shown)


modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
```diff
@@ -104,6 +99,9 @@
         this.state = new PeerState(NORMAL, 0L);
     }
 
+    ExecutorService executors =
```
aion-kelvin: Can we make executors an argument in the constructor?  Also, I don't know if using every available CPU for this task is what we want (in any case I don't think it should be decided by this class how many threads there are).

AionJayT: I think that's fine, but we can set the thread priority in the executors at the lowest level.

AionJayT: another thought, change thread priority between diffent sync mode. if it's under normal node, make priority at highest. otherwise, lowest.

modAionImpl/src/org/aion/zero/impl/sync/TaskStorePendingBlocks.java
```diff
@@ -0,0 +1,73 @@
...
+            final Logger _log) {
+        this.chain = _chain;
+        this.batch = _batch;
+        this.displayId = _displayId;
+        this.log = _log;
+    }
+
+    @Override
+    public void run() {
+        Thread.currentThread().setPriority(Thread.MAX_PRIORITY - 1);
```
AionJayT: considering the thread priority also.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
```diff
@@ -350,11 +357,11 @@ public boolean addStatusBlock(AionBlock block) {
      */
     @VisibleForTesting
     int getStatusSize() {
-        lock.readLock().lock();
+        internalLock.lock();
```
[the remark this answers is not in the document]
AlexandraRoatis: First, the methods marked by `@VisibleForTesting` are only used for testing, so any concurrency improvements are not targeted for these methods.

The concurrency improvement was targetting the decoupling the `nextBase` method which is called frequently from the rest of the database access functionality. Even a read lock acquired by this method can delay the database write operations unnecessarily.

@AionJayT is correct in saying that the objects for which access is managed with the internal lock are: `status`, `minStatus` and `maxStatus`, but `maxRequest` is also in this list. If I were to make the last variable volatile, I can no longer rely on the base values being returned incremental even when called from different threads. My intention is that there will be multiple `TaskImportBlocks` executed by parallel threads.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java
AlexandraRoatis: That is in a sense what we have there now. 
- `TaskImportBlocks` has the highest priority. 
- Saving the received blocks to disk `TaskStorePendingBlocks` has a slightly lower priority, but still high because it is part of the block importing functionality (i.e. when we reach those blocks we don't want to request them again because the task did not finish storing them). 
- Deleting the used blocks becomes critical only when the database grows very big, so the priority for `TaskDropImportedBlocks` is just slightly above normal.

modAionImpl/src/org/aion/zero/impl/db/PendingBlockStore.java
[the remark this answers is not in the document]
AionJayT: @AlexandraRoatis I am fine with your comment. It should be has a way the avoid 2 locks. but let's put this discussion as a backlog. Cause currently couldn't see issues with 2 locks.

Issue #634, opened 2018-09-12
eth_signtransaction RPC method not implemented


aion-kelvin:
Updated by @qoire and @kzeine :

# Introduction
The Web3 1.0 has a method named `eth.signTransaction`, that calls an RPC method named `eth_signTransaction`.  `eth_signTransaction` is not implemented in the kernel.  The purpose of it is to ask the kernel to sign a given transaction, but not send it (unlike `eth_sendTransaction`), so that the client can have the signed transaction to send at a later time (via `eth_sendRawTranaction`).
# Current situation
This feature is missing in the current Aion implementation. The following error occurs:
Web3 1.0 JS invocation:
```
aion-127.0.0.1:8545> eth.signTransaction({ from:'0xa0450c4333e72ed26552d7462c0b3669924eec816a219b3960d5b3f0b33f7444', to:'0xa0b88269779d225510ca880ed742e445db0c70efb1ee3159b6d56479ae3501f9', gasPrice:10000000, gas:12345, value:"1337", data:""} ).then(console.log);
Promise {
  <pending>,
  domain:
   Domain {
     domain: null,
     _events:
      { removeListener: [Function: updateExceptionCapture],
        newListener: [Function: updateExceptionCapture],
        error: [Function: debugDomainError] },
     _eventsCount: 3,
     _maxListeners: undefined,
     members: [] } }
aion-127.0.0.1:8545> (node:13250) UnhandledPromiseRejectionWarning: Error: Returned error: Method not found
    at Object.ErrorResponse (/home/sergiu/repos/aion_web3/packages/web3-core-helpers/src/errors.js:29:16)
    at /home/sergiu/repos/aion_web3/packages/web3-core-requestmanager/src/index.js:140:36
    at XMLHttpRequest.request.onreadystatechange (/home/sergiu/repos/aion_web3/packages/web3-providers-http/src/index.js:91:13)
    at XMLHttpRequestEventTarget.dispatchEvent (/home/sergiu/repos/aion_web3/packages/web3-providers-http/node_modules/xhr2-cookies/dist/xml-http-request-event-target.js:34:22)
    at XMLHttpRequest._setReadyState (/home/sergiu/repos/aion_web3/packages/web3-providers-http/node_modules/xhr2-cookies/dist/xml-http-request.js:208:14)
[...]

PR #667, opened 2018-10-11
[#634] Added eth_signTransaction implementation and tests

satran004:
## Notice

## Description

This pull request implements the missing eth.signTransaction method in kernel. This method is part of web3js 1.0.
ApiWeb3Aion.java is changed to add eth_signTransaction method.
ApiAion.java is changed to add a new method signTransaction() which is called from eth_signTransaction()

A separate pull request will be created to fix aion_web3 module for this new method.
-

Fixes Issue #634  .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing
New test cases added in ApiAionTest.java
-

## Verification

### DIFF
--- a/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
@@ -661,16 +661,7 @@ public class ApiWeb3Aion extends ApiAion {
             return new RpcMsg(
                 null, RpcError.INVALID_PARAMS, "Please check your transaction object.");
 
-        Address address = null;
-        if(_address == null || _address.trim().length() == 0)
-            return new RpcMsg(null, RpcError.INVALID_PARAMS, "Address cannot be null");
-        else
-            address = Address.wrap(_address);
-
-        ECKey key = getAccountKey(address.toString());
-        if (key == null) return new RpcMsg(null, RpcError.NOT_ALLOWED, "Account not unlocked.");
-
-        AionTransaction tx = signTransaction(txParams, key);
+        AionTransaction tx = signTransaction(txParams, _address);
         if(tx != null) {
             JSONObject obj = new JSONObject();
             obj.put("raw", TypeConverter.toJsonHex(tx.getEncoded()));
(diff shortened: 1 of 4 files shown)


modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -635,6 +635,66 @@ public RpcMsg eth_sign(Object _params) {
...
+            return new RpcMsg(
+                null, RpcError.INVALID_PARAMS, "Please check your transaction object.");
+
+        Address address = null;
+        if(_address == null || _address.trim().length() == 0)
+            return new RpcMsg(null, RpcError.INVALID_PARAMS, "Address cannot be null");
+        else
+            address = Address.wrap(_address);
+
+        ECKey key = getAccountKey(address.toString());
```
arajasek: Can you please move things like the nullity checks, and the account unlocked check to the ApiAion.signTransaction() function? 

The reason is that we also support a different API, and this would avoid code duplication. For now, simply log and error and return null from the ApiAion.signTransaction() when necessary. 

Also I'd suggest sending the address as a string to ApiAion, and doing the key extraction there.

modApiServer/test/org/aion/api/server/ApiAionTest.java
```diff
@@ -421,4 +428,85 @@ public void testHashRate() {
...
+
+        JSONArray jsonArray = new JSONArray();
+        jsonArray.put(tx);
+        jsonArray.put(addr);
+
+        RpcMsg rpcMsg = web3Api.eth_signTransaction(jsonArray);
+        assertNotNull(rpcMsg);
+
+        JSONObject result = (JSONObject)rpcMsg.getResult();
+        assertNotNull(result);
```
aion-kelvin: I think we could be a bit more stringent with this test.  I suggest adding one or both of these:

1. Verify the sub-fields within `result.get("tx")` against the original transaction that the test signed to make sure they're correct in the signed transaction
2. Verify that the signed transaction is valid by calling eth_sendRawTransaction with`result ` and seeing that it doesn't return null / throw an exception

PR #697, opened 2018-11-06
Block propagation enhancement

AlexandraRoatis:
## Description

1. Propagating blocks only if they have a valid result (defined as `!equals(INVALID_BLOCK) && !equals(CONSENSUS_BREAK)`) after import attempt.
2. Expanded logs to print the block timestamp.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing test suite

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
@@ -576,8 +576,15 @@ public class AionBlockchainImpl implements IAionBlockchain {
 
         if (block.getTimestamp()
                 > (currTimeSeconds
-                        + this.chainConfiguration.getConstants().getClockDriftBufferTime()))
+                        + this.chainConfiguration.getConstants().getClockDriftBufferTime())) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(
+                        "Block {} invalid due to timestamp {}.",
+                        block.getShortHash(),
+                        block.getTimestamp());
+            }
             return INVALID_BLOCK;
+        }
 
         if (LOG.isDebugEnabled()) {
             LOG.debug(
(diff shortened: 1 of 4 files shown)


modMcf/src/org/aion/mcf/core/ImportResult.java
```diff
@@ -46,4 +46,8 @@ public boolean isNotBest() {
     public boolean isStored() {
         return equals(EXIST) || isSuccessful();
     }
+
+    public boolean isValid() {
+        return !equals(INVALID_BLOCK) && !equals(CONSENSUS_BREAK);
```
arajasek: I would write this as

 `!(equals(INVALID_BLOCK) || equals(CONSENSUS_BREAK)`

 This is my only suggested change :P

AlexandraRoatis: do you mean that `CONSENSUS_BREAK` should be considered valid?

AionJayT: this is just the logic gate combination. very small performance enhancement :)

arajasek: Yeah, what Jay said...I would always prefer to write !(X or Y) instead of (!X and !Y).

It was only a half-serious suggestion x)

qoire: I'm not seeing ``CONSENSUS_BREAK`` used anywhere outside of tests

AlexandraRoatis: ok, will update it

AlexandraRoatis: @qoire so, you would prefer that we don't include it here?

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -576,8 +576,15 @@ ImportResult tryToConnectInternal(final AionBlock block, long currTimeSeconds) {
 
         if (block.getTimestamp()
                 > (currTimeSeconds
-                        + this.chainConfiguration.getConstants().getClockDriftBufferTime()))
+                        + this.chainConfiguration.getConstants().getClockDriftBufferTime())) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(
+                        "Block {} invalid due to timestamp {}.",
+                        block.getShortHash(),
```
qoire: Suggest using full hash here, makes everyones lives easier for a little more verbosity

Issue #698, opened 2018-11-06
Audit usage of Guava cache 


ali-sharif:
@AlexandraRoatis discovered an issue with the guava chache persisting to disk in low-memory-available situations (which still happens). It was discovered to be causing ugly race conditions with data retrieval from the database. 

The guava cache is still being used, primarily in the json rpc api implementation. 

Can we investigate: 
1) https://github.com/ben-manes/caffeine as a replacement to guava cache
2) conditions / known bugs under which guava would store cache-set to disk

The outcome here would be to have enough confidence in either a new release of Guava cache or Caffeine cache to re-enable the application-level caching features for the database layer.

arajasek:
PR #699 switches us to the Caffeine library for Web3

PR #699, opened 2018-11-07
Quickfix to prevent guava cache from persisting on disk after shutdown

arajasek:
## Notice

## Description

As discussed in Issue #698, there is a persistence-related bug with the Guava cache, which is used by the Web3 API. This quickfix destroys all caches on shutdown of the API, which mitigates the issue for now.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

All existing tests pass

## Verification


AlexandraRoatis:
in my test, the cache folder persisted with all entries even after shutdown

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2461,10 +2399,10 @@ public RpcMsg ops_getTransactionReceiptListByBlockHash(Object _params) {
         // ok to getUnchecked() since the load() implementation does not throw checked exceptions
         AionBlock b;
         try {
-            b = blockCache.getUnchecked(new ByteArrayWrapper(blockHash));
-        } catch (CacheLoader.InvalidCacheLoadException e) {
+            b = blockCache.get(new ByteArrayWrapper(blockHash));
+        } catch (Exception e) {
```
aion-kelvin: This catch got wider ... is there no equivalent more specific exception in Caffeine?

aion-kelvin: Inside the catch block, there's a comment "Catch errors if send an incorrect tx hash" and then it returns RpcError.INVALID_REQUEST, which implies the user/caller provided input that's invalid.

I think in the original code, it's catching only cases where the given tx hash does not exist.  But the new catch block seems wider than that.  Would it be possible to catch only the exception that corresponds to the tx hash not existing (i think NullPointerException, when I took a look at LoadingCache.get()'s JavaDoc), and then another catch block for other exceptions, which returns some different error (internal error or something to that effect)?

Issue #661, opened 2018-10-04
Expand Sync Statistics


AlexandraRoatis:
# Introduction
We are looking to expand the sync statistics for the Aion Kernel to produce a more comprehensive set that helps users and developers better monitor the Kernel.
# Current situation
The feature set described below is missing from the current Aion implementation.
# Task Description
We're asking for contributors to modify the class [`SyncStats.java`](https://github.com/aionnetwork/aion/blob/master-pre-merge/modAionImpl/src/org/aion/zero/impl/sync/SyncStats.java) to expand the sync statistics gathered and printed in the logs.
The functionality to be added is:
* The **percentage of requests** made to each peer with respect to the total number of requests made.
* A list of peers ordered by the **total number of blocks received** from each peer used to determine who is providing the majority of blocks, i.e. top seeds.
* A list of peers ordered by the **total number of blocks requested** by each peer used to determine who is requesting the majority of blocks, i.e. top leeches.
* **Average response time** between sending blocks out and that peer responding shown for each peer and averaged for all peers.
* **Unit tests** showing correct functionality and full coverage of the methods added to SyncStats class to provide the above-mentioned functionality additions.
# Acceptance Criteria
- The code implements the targets described above.
- The code is well documented and fully covered by unit tests.
- The code is compliant with Aion's [code conventions](https://github.com/aionnetwork/aion/wiki/Aion-Code-Conventions).
- The code gets merged into the main code repository.
# Required Skill
Participants need to have experience in Java software development, debugging and have basic knowledge about blockchain systems.
# Guidance
Details regarding this issue can be requested in comments.
Further technical support is available through [Gitter](https://gitter.im/aionnetwork/Lobby).

AlexandraRoatis:
@santteegt you are right, some deep debugging will be needed to complete the task. You should be able to import the project using Gradle in your IDE of choice. Changes must be based on the [`master-pre-merge`](https://github.com/aionnetwork/aion/tree/master-pre-merge) branch. The data for the statistics will be collected from [`TaskImportBlocks.java`](https://github.com/aionnetwork/aion/blob/master-pre-merge/modAionImpl/src/org/aion/zero/impl/sync/TaskImportBlocks.java) and the information should be printed in [`TaskShowStatus.java`](https://github.com/aionnetwork/aion/blob/master-pre-merge/modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java) under debug logging. Both classes already reference a common `SyncStats` object.

[...]

PR #700, opened 2018-11-09
Enabling choice of stats display from config file

AlexandraRoatis:
## Description

I updated the configuration file to allow the choice of individual stats to be displayed during execution.

Expands on Issue #661.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing test suite

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
+++ b/modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
@@ -25,10 +25,12 @@ package org.aion.zero.impl.sync;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.aion.base.util.Hex;
+import org.aion.mcf.config.StatsType;
 import org.aion.p2p.INode;
 import org.aion.p2p.IP2pMgr;
 import org.aion.zero.impl.AionBlockchainImpl;
@@ -57,6 +59,7 @@ final class TaskShowStatus implements Runnable {
     private final IP2pMgr p2p;
 
     private final Map<Integer, PeerState> peerStates;
+    private final List<StatsType> showStatistics;
 
     TaskShowStatus(
             final AtomicBoolean _start,
@@ -66,6 +69,7 @@ final class TaskShowStatus implements Runnable {
             final SyncStats _stats,
             final IP2pMgr _p2p,
             final Map<Integer, PeerState> _peerStates,
+            final List<StatsType> showStatistics,
             final Logger _log) {
         this.start = _start;
         this.interval = _interval;
@@ -75,6 +79,7 @@ final class TaskShowStatus implements Runnable {
         this.p2p = _p2p;
         this.peerStates = _peerStates;
         this.p2pLOG = _log;
+        this.showStatistics = Collections.unmodifiableList(showStatistics);
     }
 
     @Override
@@ -105,11 +110,40 @@ final class TaskShowStatus implements Runnable {
 
             p2pLOG.info(status);
 
-            if (p2pLOG.isDebugEnabled()) {
-                String s = dumpPeerStateInfo(p2p.getActiveNodes().values());
-                s += dumpPeerStatsInfo();
-                if (!s.isEmpty()) {
-                    p2pLOG.debug(s);
+            String requestedInfo;
+
+            if (showStatistics.contains(StatsType.PEER_STATES)) {
+                requestedInfo = dumpPeerStateInfo(p2p.getActiveNodes().values());
+                if (!requestedInfo.isEmpty()) {
+                    p2pLOG.info(requestedInfo);
+                }
+            }
+
+            if (showStatistics.contains(StatsType.REQUESTS)) {
+                requestedInfo = dumpRequestsInfo();
+                if (!requestedInfo.isEmpty()) {
+                    p2pLOG.info(requestedInfo);
+                }
+            }
+
+            if (showStatistics.contains(StatsType.SEEDS)) {
+                requestedInfo = dumpTopSeedsInfo();
+                if (!requestedInfo.isEmpty()) {
+                    p2pLOG.info(requestedInfo);
+                }
+            }
+
+            if (showStatistics.contains(StatsType.LEECHES)) {
+                requestedInfo = dumpTopLeechesInfo();
+                if (!requestedInfo.isEmpty()) {
+                    p2pLOG.info(requestedInfo);
+                }
+            }
+
+            if (showStatistics.contains(StatsType.RESPONSES)) {
+                requestedInfo = dumpResponseInfo();
+                if (!requestedInfo.isEmpty()) {
+                    p2pLOG.info(requestedInfo);
                 }
             }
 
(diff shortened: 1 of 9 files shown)


modMcf/src/org/aion/mcf/config/CfgSync.java
```diff
@@ -38,13 +41,14 @@
...
     private boolean showStatus;
+    private List<StatsType> showStatistics;
 
     private static int BLOCKS_QUEUE_MAX = 32;
 
     public CfgSync() {
         this.blocksQueueMax = BLOCKS_QUEUE_MAX;
-
         this.showStatus = false;
+        this.showStatistics = new ArrayList<>(Arrays.asList(StatsType.NONE));
```
aion-kelvin: why not represent "NONE" with just an empty list?

modMcf/src/org/aion/mcf/config/CfgSync.java
```diff
@@ -72,6 +79,36 @@ public void fromXML(final XMLStreamReader sr) throws XMLStreamException {
...
 
+    private static void parseSelectedStats(List<StatsType> showStatistics, String readValue) {
+        showStatistics.clear();
+
+        String[] selected = readValue.split(",");
+
+        for (String option : selected) {
+            try {
+                showStatistics.add(StatsType.valueOf(option.toUpperCase()));
+            } catch (Exception e) {
```
aion-kelvin: is this to ignore options that are invalid?  If so I think this should only catch IllegalArgumentException  

(according to https://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#valueOf(java.lang.Class,%20java.lang.String))

modMcf/src/org/aion/mcf/config/CfgSync.java
```diff
@@ -72,6 +79,36 @@ public void fromXML(final XMLStreamReader sr) throws XMLStreamException {
         }
     }
 
+    private static void parseSelectedStats(List<StatsType> showStatistics, String readValue) {
```
aion-kelvin: minor/optional suggestion: I think things would be simpler if showStatistics was a Set instead of a List

modMcf/src/org/aion/mcf/config/StatsType.java
```diff
@@ -0,0 +1,56 @@
...
+ * @author Alexandra Roatis
+ */
+public enum StatsType {
+    ALL,
+    PEER_STATES,
+    REQUESTS,
+    SEEDS,
+    LEECHES,
+    RESPONSES,
+    NONE; // used as default for invalid settings
```
aion-kelvin: I think the parsing logic would be simpler if NONE was represented by an empty list/set in the Java side and a blank value in the XML side

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
```diff
@@ -55,179 +54,249 @@
...
             p2pLOG.info(status);
 
-            if (p2pLOG.isDebugEnabled()) {
-                String s = dumpPeerStateInfo(p2p.getActiveNodes().values());
-                s += dumpPeerStatsInfo();
-                if (!s.isEmpty()) {
-                    p2pLOG.debug(s);
+            if (showStatistics.contains(StatsType.PEER_STATES)) {
+                requestedInfo = dumpPeerStateInfo(p2p.getActiveNodes().values());
+                if (!requestedInfo.isEmpty()) {
```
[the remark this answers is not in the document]
AlexandraRoatis: I don't think there's much value in it. One can see the full list of active peers in the p2p-status. It's easy to reason that anything in the p2p list and not displayed has no stats data.

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
```diff
@@ -55,179 +54,249 @@
...
             try {
                 Thread.sleep(interval);
             } catch (InterruptedException e) {
-                if (p2pLOG.isDebugEnabled()) {
+                // without requested shutdown
+                if (start.get() && p2pLOG.isDebugEnabled()) {
                     p2pLOG.debug("sync-ss shutdown.", e);
                 }
-                return;
+                continue;
```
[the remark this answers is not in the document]
AlexandraRoatis: The thread will be killed if the start value is set to false.

modMcf/src/org/aion/mcf/config/CfgSync.java
AlexandraRoatis: It's a possibility, but I'm not sure how user friendly that is. It definitely makes the comment in the config file longer. Up for debate though, if anyone else wants to pitch in.

modMcf/src/org/aion/mcf/config/CfgSync.java
AlexandraRoatis: why not include the `NullPointerException`s?

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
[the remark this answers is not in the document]
AlexandraRoatis: Plus the main objective here was to reduce unnecessary info in the logs.

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
```diff
@@ -55,179 +54,249 @@
...
         this.interval = _interval;
         this.chain = _chain;
         this.networkStatus = _networkStatus;
         this.stats = _stats;
-        this.printReport = _printReport;
-        this.reportFolder = _reportFolder;
         this.p2p = _p2p;
         this.peerStates = _peerStates;
         this.p2pLOG = _log;
+        this.showStatistics = Collections.unmodifiableList(showStatistics);
```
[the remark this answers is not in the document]
AlexandraRoatis: Good point. I will update it.

modMcf/src/org/aion/mcf/config/CfgSync.java
arajasek: I like making NONE an explicit option, I think that is better for code maintenance / readability, and there's no (real) performance hit.

modMcf/src/org/aion/mcf/config/CfgSync.java
arajasek: I agree with this

modMcf/src/org/aion/mcf/config/CfgSync.java
```diff
@@ -72,6 +79,36 @@ public void fromXML(final XMLStreamReader sr) throws XMLStreamException {
...
+        showStatistics.clear();
+
+        String[] selected = readValue.split(",");
+
+        for (String option : selected) {
+            try {
+                showStatistics.add(StatsType.valueOf(option.toUpperCase()));
+            } catch (Exception e) {
+                // skip option
+                continue;
```
arajasek: You could just have an empty catch block here

modAionImpl/src/org/aion/zero/impl/sync/TaskShowStatus.java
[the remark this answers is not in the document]
arajasek: `continue` can be omitted if you want

modMcf/src/org/aion/mcf/config/CfgSync.java
aion-kelvin: true, wouldn't hurt to also catch NullPointerException

PR #702, opened 2018-11-09
Compact the state after slow block imports

AlexandraRoatis:
## Description

The new sync strategy puts more pressure on the db by importing blocks significantly faster. To counterbalance this I've added functionality to call LevelDBs compaction process when block import times reach 100 milliseconds, which results in a speedup of subsequent imports.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing test suite

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
@@ -544,6 +544,10 @@ public class AionBlockchainImpl implements IAionBlockchain {
         return tryToConnectInternal(block, System.currentTimeMillis() / THOUSAND_MS);
     }
 
+    public synchronized void compactState() {
+        repository.compactState();
+    }
+
     /**
      * Processes a new block and potentially appends it to the blockchain, thereby changing the
      * state of the world. Decoupled from wrapper function {@link #tryToConnect(AionBlock)} so we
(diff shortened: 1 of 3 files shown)


sdesmond46:
> I'm seeing some issues on my full node that I think is related to this change. Whenever we get a single import which takes more than 1 second, the compaction gets triggered. While the compaction is running, the node cannot handle any API requests are do anything else but wait for the compaction to finish, which is taking 15+ minutes for me. This is getting triggered a bunch right now because of some [load testing](https://mastery.aion.network/#/charts/TransactionsoverTime) on the mastery network.
> 
> Since this effectively shuts down the kernel from doing anything useful, this behavior should be configurable and disableable. Then it would be on the kernel operator to occasionally stop the node and run `aion.sh --db-compact` before restarting it.
> 
> Here's the DB section of my config:
> 
> ```
>       <db>
>         
>         <path>database</path>
>         
>         <check_integrity>false</check_integrity>
>         
>         
>         
>         
>         <state-storage>FULL</state-storage>
>         
>         
>         <vendor>leveldb</vendor>
>         
>         <enable_db_compression>true</enable_db_compression>
>       </db>
> ```

Here's a snippet of the logs before the compacting and the result while compacting is going on (compacting starts on line 333)
[compaction_snippet.log](https://github.com/aionnetwork/aion/files/2721767/compaction_snippet.log)

PR #704, opened 2018-11-12
Included the ip and port in computing the id hash

AlexandraRoatis:
## Description

Analyzing some execution logs I noticed that if multiple peers have the same id all but the first get rejected. This restriction is unnecessary and leads to a random selection in which peer gets to connect to a node.

This PR updates the computed id hash to include the IP and port number, ensuring unique kernel instances are considered separately as peers. Effectively it allows that if kernel A and kernel B have the same id, then a kernel C (using this implementation) can communicate with both A and B, whereas before it would have selected the first node it interacted with and rejected the other.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing test suite + a few specific tests

## Verification


AlexandraRoatis:
Closing until adding two more features:
1. Update the display ID to show that the two nodes are distinct.
2. Ensure that kernel A and kernel B can also communicate with each other.

PR #705, opened 2018-11-14
P2p connection drop fix

AionJayT:
## Notice

## Description

- fix p2p tempnodelist add/remove potential locking issue.
- remove redundant thread pool in TaskSend
- add time check to prevent the channelsocket write the message but buffer is full cause drop connection.

Fixes Issue # .

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

- passed the existing p2p unit test

## Verification

### DIFF
--- a/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
+++ b/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
@@ -164,8 +164,10 @@ public class NodeMgr implements INodeMgr {
 
     /** @param _n Node */
     @Override
-    public void addTempNode(final INode _n) {
-        tempNodesLock.lock();
+    public void addTempNode(final INode _n) throws InterruptedException {
+
+        tempNodesLock.lockInterruptibly();
+
         try {
             if (tempNodes.size() < maxTempNodes
                     && !tempNodes.containsKey(_n.getPeerId())
@@ -196,13 +198,19 @@ public class NodeMgr implements INodeMgr {
     }
 
     @Override
-    public INode tempNodesTake() {
+    public INode tempNodesTake() throws InterruptedException {
         INode node = null;
-        tempNodesLock.lock();
+
+        tempNodesLock.lockInterruptibly();
+
+        if (tempNodes.isEmpty()) {
+            return null;
+        }
+
         try {
             node = tempNodes.remove(tempNodes.keySet().iterator().next());
         } catch (Exception e) {
-            p2pLOG.error("<tempNodesTake IllegalStateException>", e);
+            p2pLOG.error("<tempNodesTake Exception>", e);
         } finally {
             tempNodesLock.unlock();
         }
--- a/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
+++ b/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
@@ -433,10 +433,6 @@ public class NodeMgr implements INodeMgr {
 
     public void dropActive(int nodeIdHash, String _reason) {
 
-        if (p2pLOG.isDebugEnabled()) {
-            p2pLOG.debug("<dropActive idHash:{} reason:{}>", nodeIdHash, _reason);
-        }
-
         INode node = null;
         try {
             node = activeNodes.remove(nodeIdHash);
--- a/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
+++ b/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
@@ -166,6 +166,8 @@ public class NodeMgr implements INodeMgr {
     @Override
     public void addTempNode(final INode _n) throws InterruptedException {
 
+        if (_n == null) return;
+
         tempNodesLock.lockInterruptibly();
 
         try {
--- a/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
+++ b/modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
@@ -36,7 +36,6 @@ import java.util.OptionalDouble;
 import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 import org.aion.p2p.INode;
 import org.aion.p2p.INodeMgr;
@@ -54,9 +53,7 @@ public class NodeMgr implements INodeMgr {
     private final int maxTempNodes;
     private final Set<String> seedIps = new HashSet<>();
     private final IP2pMgr p2pMgr;
-    private final ReentrantLock takeLock = new ReentrantLock();
-    private final ReentrantLock putLock = new ReentrantLock();
-    private final Condition notEmpty = takeLock.newCondition();
+    private final ReentrantLock tempNodesLock = new ReentrantLock();
     private final Map<Integer, INode> tempNodes =
             Collections.synchronizedMap(new LinkedHashMap<>());
     private final Map<Integer, INode> outboundNodes = new ConcurrentHashMap<>();
@@ -168,20 +165,17 @@ public class NodeMgr implements INodeMgr {
     /** @param _n Node */
     @Override
     public void addTempNode(final INode _n) {
-        final ReentrantLock putLock = this.putLock;
+        tempNodesLock.lock();
         try {
-            putLock.lockInterruptibly();
-
             if (tempNodes.size() < maxTempNodes
                     && !tempNodes.containsKey(_n.getPeerId())
                     && (notActiveNode(_n.getIdHash()) || _n.getIfFromBootList())) {
                 tempNodes.putIfAbsent(_n.getPeerId(), _n);
-                signalNotEmpty();
             }
-        } catch (InterruptedException e) {
+        } catch (Exception e) {
             p2pLOG.error("<addTempNode exception>", e);
         } finally {
-            putLock.unlock();
+            tempNodesLock.unlock();
         }
     }
 
@@ -204,22 +198,13 @@ public class NodeMgr implements INodeMgr {
     @Override
     public INode tempNodesTake() {
         INode node = null;
-        final ReentrantLock takeLock = this.takeLock;
+        tempNodesLock.lock();
         try {
-            takeLock.lockInterruptibly();
-            while (tempNodes.isEmpty()) {
-                notEmpty.await();
-            }
-
-            Integer key = tempNodes.keySet().iterator().next();
-            node = tempNodes.remove(key);
-            if (!tempNodes.isEmpty()) {
-                notEmpty.signal();
-            }
-        } catch (InterruptedException e) {
+            node = tempNodes.remove(tempNodes.keySet().iterator().next());
+        } catch (Exception e) {
             p2pLOG.error("<tempNodesTake IllegalStateException>", e);
         } finally {
-            takeLock.unlock();
+            tempNodesLock.unlock();
         }
 
         return node;
@@ -515,16 +500,6 @@ public class NodeMgr implements INodeMgr {
             }
         } catch (NullPointerException e) {
             p2pLOG.info("<p2p-ban null exception>", e);
-        }
-    }
-
-    private void signalNotEmpty() {
-        final ReentrantLock takeLock = this.takeLock;
-        takeLock.lock();
-        try {
-            notEmpty.signal();
-        } finally {
-            takeLock.unlock();
         }
     }
 }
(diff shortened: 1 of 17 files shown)


modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
```diff
@@ -204,22 +198,13 @@ public void addOutboundNode(final INode _n) {
...
-                notEmpty.await();
-            }
-
-            Integer key = tempNodes.keySet().iterator().next();
-            node = tempNodes.remove(key);
-            if (!tempNodes.isEmpty()) {
-                notEmpty.signal();
-            }
-        } catch (InterruptedException e) {
+            node = tempNodes.remove(tempNodes.keySet().iterator().next());
```
AlexandraRoatis: This will trigger a NoSuchElementException when the iterator is empty. I think this exception should be avoided with proper if statements checking that there are elements in the set instead of logging the problem. These exceptions will make debugging confusing instead of helping.

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
```diff
@@ -168,20 +165,17 @@ public boolean isSeedIp(String _ip) {
     /** @param _n Node */
     @Override
     public void addTempNode(final INode _n) {
-        final ReentrantLock putLock = this.putLock;
+        tempNodesLock.lock();
         try {
-            putLock.lockInterruptibly();
-
             if (tempNodes.size() < maxTempNodes
                     && !tempNodes.containsKey(_n.getPeerId())
```
AlexandraRoatis: Please update the JavaDoc to say that `_n` cannot be null.

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -86,6 +86,8 @@
 
     public static final Logger p2pLOG = AionLoggerFactory.getLogger(LogEnum.P2P.name());
 
+    public static final int WORKER = 32;
```
AlexandraRoatis: Why 32?

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -198,15 +201,13 @@ public void run() {
                         });
             }
 
-            int pNum = Runtime.getRuntime().availableProcessors();
-
-            for (int i = 0; i < (pNum << 1); i++) {
+            for (int i = 0; i < WORKER; i++) {
```
AlexandraRoatis: the number of send processes used to be double the number of receive processes, what's the meaning behind having them equal now?

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -333,6 +334,7 @@ public void closeSocket(SocketChannel _sc, String _reason, Exception e) {
             SelectionKey sk = _sc.keyFor(selector);
             if (sk != null) {
                 sk.cancel();
+                sk.attach(null);
```
AlexandraRoatis: is this done to improve garbage collection by reducing reference counting to the previous object?

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskInbound.java
```diff
@@ -155,12 +153,11 @@ public void run() {
...
     }
 
-    private void accept() throws Exception {
+    private void accept(SelectableChannel _channel) throws Exception {
         if (this.nodeMgr.activeNodesSize() >= this.mgr.getMaxActiveNodes()) {
             return;
         }
-
-        SocketChannel channel = this.tcpServer.accept();
+        SocketChannel channel = ((ServerSocketChannel) _channel).accept();
```
AlexandraRoatis: I feel like there should be an instance-of check before doing the casting. Perhaps the method should receive a `ServerSocketChannel` parameter and the calling method above can do the check and casting.

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskSend.java
```diff
@@ -67,17 +63,14 @@ public TaskSend(
...
-                            TOTAL_LANE,
-                            0,
-                            TimeUnit.MILLISECONDS,
-                            new LinkedBlockingQueue<>(THREAD_Q_LIMIT),
-                            Executors.defaultThreadFactory());
-        }
+        this.tpe =
+                new ThreadPoolExecutor(
+                        1,
+                        1,
```
AlexandraRoatis: will this change impact performance?

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskWrite.java
```diff
@@ -92,25 +92,50 @@ public void run() {
...
+                        // without sleep will hangup this thread.
+                        Thread.sleep(0, 1);
+                    }
+
+                    t2 = System.currentTimeMillis() - t1;
+                } while (buf.hasRemaining() && (t2 < 100));
+
+                if (p2pLOG.isTraceEnabled() && (t2 > 10)) {
+                    p2pLOG.trace(
+                            "msg write: id {} size {} time {}ms length {}",
```
AlexandraRoatis: minor update: `time {} ms` instead of `time {}ms`

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
```diff
@@ -168,20 +165,17 @@ public boolean isSeedIp(String _ip) {
     /** @param _n Node */
     @Override
     public void addTempNode(final INode _n) {
-        final ReentrantLock putLock = this.putLock;
+        tempNodesLock.lock();
```
aion-kelvin: Previously, it used lockInterruptibly() instead of lock().  Should we also do lockInterruptibly here?

I'm not that familiar with the difference, but from the JavaDoc, it seems when the thread blocks on lock(), it can't be interrupted until the lock is acquired.  Not sure if this will do something strange if the kernel shuts down while this thread is blocked on lock()

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
```diff
@@ -204,22 +198,13 @@ public void addOutboundNode(final INode _n) {
...
-            if (!tempNodes.isEmpty()) {
-                notEmpty.signal();
-            }
-        } catch (InterruptedException e) {
+            node = tempNodes.remove(tempNodes.keySet().iterator().next());
+        } catch (Exception e) {
             p2pLOG.error("<tempNodesTake IllegalStateException>", e);
         } finally {
-            takeLock.unlock();
+            tempNodesLock.unlock();
```
aion-kelvin: Same question about lock vs lockInterrruptibly

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
aion-kelvin: Indeed, I also think there needs to be an empty check before calling tempNodes.remove.

When I looked at the code in master-pre-merge, only TaskconnectPeers.run calls this method and it handles null properly.  So maybe one way to fix this is just: if(tempNodes.isEmpty()) return null;

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
```diff
@@ -204,22 +198,13 @@ public void addOutboundNode(final INode _n) {
...
-
-            Integer key = tempNodes.keySet().iterator().next();
-            node = tempNodes.remove(key);
-            if (!tempNodes.isEmpty()) {
-                notEmpty.signal();
-            }
-        } catch (InterruptedException e) {
+            node = tempNodes.remove(tempNodes.keySet().iterator().next());
+        } catch (Exception e) {
             p2pLOG.error("<tempNodesTake IllegalStateException>", e);
```
aion-kelvin: Log message is different from the actual exception being caught

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -176,7 +178,8 @@ public void run() {
             tcpServer = ServerSocketChannel.open();
             tcpServer.configureBlocking(false);
             tcpServer.socket().setReuseAddress(true);
-            tcpServer.socket().bind(new InetSocketAddress(Node.ipBytesToStr(selfIp), selfPort));
+            tcpServer.socket().setReceiveBufferSize(128*1024);
```
aion-kelvin: a comment explaining the number would be helpful

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskSend.java
aion-kelvin: I'm also curious about the performance impact.  Actually, why do we need to change this from TOTAL_LANE to 1?

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskWrite.java
```diff
@@ -92,25 +92,50 @@ public void run() {
...
+                    int result = sc.write(buf);
+                    wrote += result;
+
+                    if (result == 0) {
+                        // @Attention:  very important sleep , otherwise when NIO write buffer full,
+                        // without sleep will hangup this thread.
+                        Thread.sleep(0, 1);
+                    }
+
+                    t2 = System.currentTimeMillis() - t1;
```
aion-kelvin: It may be slightly safer to use System.nanoTime() for this measurement instead of System.currentTimeMillis().

curentTimeMillis changes if the system's clock changes, whereas nanoTime() does not: "This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time."

https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
AionJayT: just see some guidance recommend to clean the buffer in case the gc is not smart.

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskSend.java
AionJayT: Before bug fix, the Thread executor was shared with whole Tasksend thread. But in this PR, each Tasksend thread has one dedicated thread to process the write message. (please see the runnable submit in the thread)

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
AionJayT: yap, the lockInterruptibly it a little bit better in this scenario.

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
AionJayT: it's by the design of the method TaskSend.hash2Lane()

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
AionJayT: add empty check

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
AionJayT: added null check and will return directly

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
AionJayT: each worker thread represents a lane to deal with the peer connection belong to this lane. Cause now the lane is fixed number, then the thread number should be fixed.

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskWrite.java
AionJayT: updated

modP2pImpl/src/org/aion/p2p/impl/comm/NodeMgr.java
AionJayT: updated

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
AionJayT: updated

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskWrite.java
AionJayT: updated

PR #731, opened 2018-11-28
Mongo

sdesmond46:
## Notice

## Description

This change adds support for a new database vendor - MongoDB. This will be the first database which is not stored in the file system where the kernel is running, which allows for more flexibility when deploying and running kernels.

The change introduces a runtime dependency on the Mongo Java Driver, and a test dependency on Mongo DB itself so we can test against a localhost Mongo instance.

To use the Mongo driver to connect to a Mongo database running on your local machine, you would update the `config.xml` with `<vendor>mongodb</vendor>` and `<path>mongodb://localhost:27017</path>`

## Type of change

- [ ] Bug fix.
- [X] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

- Tested running the kernel locally with Mongo and with the existing db providers
- Added unit tests
- These changes are actively running on the [Nodesmith](https://docs.nodesmith.io/#/) service without issue

## Verification

- [X] I have self-reviewed my own code and conformed to the style guidelines of this project.
- [X] New and existing tests pass locally with my changes.
- [X] I have added tests for my fix or feature.
- [X] I have made appropriate changes to the corresponding documentation.
- [X] My code generates no new warnings.
- [X] Any dependent changes have been made.


AionJayT:
@sdesmond46 Thanks for your contribution. Do you have benchmark test or data for the implementation?

sdesmond46:
> @sdesmond46 Thanks for your contribution. Do you have benchmark test or data for the implementation?

I didn't do any formal benchmark testing on it, but I didn't see any real performance issues compared to one LevelDB when Mongo was running on the same machine. There is obviously a bit more latency when Mongo is running on a different machine, but it certainly wasn't something that made the kernel unusable.

I'm happy to add some more formal testing if that would help though.

AionJayT:
> @AionJayT thanks for taking a look!
> 
> Couple of responses to your questions:
> 
> 1. Good call, I bumped some of the levels down on some logs
> 2. I looked into this and couldn't quite figure out an elegant solution. Is the issue that the file is too large to be included as a test resource, or the file is too large to be in the git repo? I can dig into including & building the source, but I think the Mongo repo is pretty large
cause usually we don't put that big file into the github repo. The better way I think is writing a script to build the mongo db when the developer tries to run the mongo db test in the first time.

> 3. The license change is totally fine with me :)
> 4. I noticed that error too and made a change to suppress it. It was just a place where we were checking if the file was saved to disk instead of checking that it is persisted somewhere.

script/prepack.sh
```diff
@@ -33,7 +33,7 @@ if [ "$noGui" != "true" ] && [ ! -d "$JAVAFX_PATH" ]; then
 fi
 
 module_path=$JDK_PATH/jmods 
-add_modules="java.base,java.xml,java.logging,java.management,jdk.unsupported,jdk.sctp"
+add_modules="java.base,java.xml,java.logging,java.management,jdk.unsupported,jdk.sctp,java.security.sasl"
```
AlexandraRoatis: @aion-kelvin I think this change will impact the GUI PR. Just making a note here to not lose track of the change.

sdesmond46:
@aion-kelvin - Re: using Docker and the Mongo binaries -- The script I added runs mongo with the docker image available on DockerHub, `docker run` will pull the Docker image from Docker Hub if it doesn't exist on the local machine. I used docker for local testing since it cleans its self up easier and works no matter what OS I'm using.

The `mongod` and `mongo` binaries are added are only used in the unit tests, since I didn't think it would be ok to require Docker to be installed on running to have the tests pass. @AionJayT would it be reasonable to require Docker for the tests? Then we wouldn't need to check in the binaries I did -- when the tests ran the docker image would just get pulled from docker hub if it didn't exist on the local machine. I'll also look into adding a script to build mongo from source in the meantime

aion-kelvin:
Thanks for the changes!  Code looks good to me -- once we figure out how to package the mongo binary I think we're good to go.

> The `mongod` and `mongo` binaries are added are only used in the unit tests, since I didn't think it would be ok to require Docker to be installed on running to have the tests pass. @AionJayT would it be reasonable to require Docker for the tests? Then we wouldn't need to check in the binaries I did -- when the tests ran the docker image would just get pulled from docker hub if it didn't exist on the local machine. I'll also look into adding a script to build mongo from source in the meantime.

I'll defer to @AionJayT for the final decision, but IMO if the test needs to call an external program, it is an integration test, not a unit test.  And I think it is ok for an integration test to need extra set-up.

AionJayT:
@sdesmond46 I total agree with using docker for testing the mongo. Could you able to a this change?

sdesmond46:
> @sdesmond46 I total agree with using docker for testing the mongo. Could you able to do this change?

Yep, I just made the changes. They assume that there is a running docker daemon on the machine, and exec and cleanup the mongo container, but remove the mongo binaries from the repo

AionJayT:
@sdesmond46 the ciBuild has tailed tests, could you able to fix it?
org.aion.db.impl.AccessWithExceptionTest > initializationError FAILED
    java.lang.RuntimeException
        Caused by: java.lang.reflect.InvocationTargetException
            Caused by: java.lang.AssertionError at AccessWithExceptionTest.java:86

org.aion.db.impl.DriverBaseTest > initializationError FAILED
    java.lang.NoClassDefFoundError at DriverBaseTest.java:307

org.aion.db.impl.ConcurrencyTest > initializationError FAILED
    java.lang.RuntimeException
        Caused by: java.lang.reflect.InvocationTargetException
            Caused by: java.lang.NoClassDefFoundError at ConcurrencyTest.java:90

sdesmond46:
> @sdesmond46 the ciBuild has tailed tests, could you able to fix it?
> org.aion.db.impl.AccessWithExceptionTest > initializationError FAILED
> java.lang.RuntimeException
> Caused by: java.lang.reflect.InvocationTargetException
> Caused by: java.lang.AssertionError at AccessWithExceptionTest.java:86
> 
> org.aion.db.impl.DriverBaseTest > initializationError FAILED
> java.lang.NoClassDefFoundError at DriverBaseTest.java:307
> 
> org.aion.db.impl.ConcurrencyTest > initializationError FAILED
> java.lang.RuntimeException
> Caused by: java.lang.reflect.InvocationTargetException
> Caused by: java.lang.NoClassDefFoundError at ConcurrencyTest.java:90

Yeah .I spent some time digging it this and trying to figure out a solution,. The root cause is the tests can't connect to a running Docker service to start up the mongo container when running in the CI environment. I'm not super familiar with Jenkins and the way the runners are set up for Aion, so I wasn't sure if that's because there isn't a docker service running, or I don't know the right DOCKER_HOST to connect to it. Is there any info on how the Jenkins agents are configured for the Aion CI environment? Thanks for the help!

AionJayT:
merge this PR in and will try (or ignore) to fix the fail tests.

Issue #725, opened 2018-11-26
Port number as CLI input


AlexandraRoatis:
A new feature that would allow running the kernel with a given port number.

For example, `./aion.sh --port 12345` will start a kernel that uses port `12345`. This value should be saved in the config file for the kernel execution directory.

The option should work correctly in combination with the `-n` and `-d` options.

PR #736, opened 2018-11-29
Cli port input

beidouz:
## Notice

## Description

-  Allow running the kernel with a given port number. 

Fixes Issue #725 .

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
+++ b/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
@@ -191,6 +191,37 @@ public class Cli {
             // true means the UUID must be set
             boolean overwrite = cfg.fromXML(configFile);
 
+            // port option can be used with the -n, -d, -c, -i arguments
+            if (options.getPort() != null) {
+
+                int currentPort = cfg.getNet().getP2p().getPort();
+                int portNumber = currentPort;
+                boolean validPort = true;
+
+                try {
+                    portNumber = Integer.parseInt(options.getPort());
+                } catch (NumberFormatException e) {
+                    validPort = false;
+                    System.out.println("Port must be an integer value");
+                }
+
+                if (portNumber < 0 || portNumber > 0xFFFF) {
+                    validPort = false;
+                    System.out.println("Port out of range: " + portNumber);
+                }
+
+                if (validPort && portNumber != currentPort) {
+                    // update port in config
+                    cfg.getNet().getP2p().setPort(portNumber);
+                    overwrite = true;
+                    System.out.println("Port set to: " + portNumber);
+                } else {
+                    System.out.println(
+                        "Using the current port configuration: " + currentPort);
+                }
+                // no return, allow for other parameters combined with -c, -p
+            }
+
             // 4. can be influenced by the -d argument above
 
             if (options.getConfig() != null) {
@@ -230,36 +261,6 @@ public class Cli {
             }
 
             // 5. options that can be influenced by the -d and -n arguments
-
-            if (options.getPort() != null) {
-
-                int currentPort = cfg.getNet().getP2p().getPort();
-                int portNumber = currentPort;
-                boolean validPort = true;
-
-                try {
-                    portNumber = Integer.parseInt(options.getPort());
-                } catch (NumberFormatException e) {
-                    validPort = false;
-                    System.out.println("Port must be an integer value");
-                }
-
-                if (portNumber < 0 || portNumber > 0xFFFF) {
-                    validPort = false;
-                    System.out.println("Port out of range: " + portNumber);
-                }
-
-                if (validPort && portNumber != currentPort) {
-                    // update port in config
-                    cfg.getNet().getP2p().setPort(portNumber);
-                    overwrite = true;
-                    System.out.println("Port set to: " + portNumber);
-                } else {
-                    System.out.println(
-                        "Using the current port configuration: " + currentPort);
-                }
-                // no return, allow for other parameters combined with -p
-            }
 
             if (options.isInfo()) {
                 System.out.println(
--- a/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
+++ b/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
@@ -231,6 +231,12 @@ public class Cli {
 
             // 5. options that can be influenced by the -d and -n arguments
 
+            if (options.getPort() != null) {
+                // update port in the config file
+                cfg.toXML(new String[] {"--p2p=" + ","+ options.getPort()}, configFile);
+                // no return, allow for other parameters combined with -p
+            }
+
             if (options.isInfo()) {
                 System.out.println(
                         "Reading config file from: "
@@ -860,6 +866,9 @@ public class Cli {
             }
             if (options.getDirectory() != null) {
                 skippedTasks.add("--datadir");
+            }
+            if (options.getPort() != null) {
+                skippedTasks.add("--port");
             }
             if (options.getConfig() != null) {
                 skippedTasks.add("--config");
--- a/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
+++ b/modAionImpl/src/org/aion/zero/impl/cli/Cli.java
@@ -232,15 +232,31 @@ public class Cli {
             // 5. options that can be influenced by the -d and -n arguments
 
             if (options.getPort() != null) {
-                int port_number = Integer.parseInt(options.getPort());
-                if (port_number < 0 || port_number > 0xFFFF) {
-                    System.out.println(
-                            "\nport out of range: "
-                                    + port_number
-                                    + ", switching to the current port configuration: \n");
+
+                int currentPort = cfg.getNet().getP2p().getPort();
+                int portNumber = currentPort;
+                boolean validPort = true;
+
+                try {
+                    portNumber = Integer.parseInt(options.getPort());
+                } catch (NumberFormatException e) {
+                    validPort = false;
+                    System.out.println("Port must be an integer value");
+                }
+
+                if (portNumber < 0 || portNumber > 0xFFFF) {
+                    validPort = false;
+                    System.out.println("Port out of range: " + portNumber);
+                }
+
+                if (validPort && portNumber != currentPort) {
+                    // update port in config
+                    cfg.getNet().getP2p().setPort(portNumber);
+                    overwrite = true;
+                    System.out.println("Port set to: " + portNumber);
                 } else {
-                    // update port in the config file
-                    cfg.toXML(new String[] {"--p2p=" + "," + options.getPort()}, configFile);
+                    System.out.println(
+                        "Using the current port configuration: " + currentPort);
                 }
                 // no return, allow for other parameters combined with -p
             }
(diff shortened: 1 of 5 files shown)


modAionImpl/src/org/aion/zero/impl/config/CfgAion.java
```diff
@@ -354,7 +354,7 @@ public void toXML(final String[] args, File file) {
                     override = true;
                     String[] subArgsArr = arg.replace("--p2p=", "").split(",");
                     if (subArgsArr.length == 2) {
-                        this.getNet().getP2p().setIp(subArgsArr[0]);
+                        if (!subArgsArr[0].equals("")) this.getNet().getP2p().setIp(subArgsArr[0]);
```
aion-kelvin: Could you run the code formatter with our formatting guidelines?  I think we put the code of the if inside { } or at least on a different line.

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -226,6 +226,12 @@ public ReturnType call(final String[] args, Cfg cfg) {
 
             // 5. options that can be influenced by the -d and -n arguments
 
+            if (options.getPort() != null) {
+                // update port in the config file
+                cfg.toXML(new String[] {"--p2p=" + ","+ options.getPort()}, configFile);
```
aion-kelvin: I noticed you add a `,` to the string, then in CfgAion.java, where it is used, you split on the `,` character.  Since this parameter takes only a single argument, it would be more intuitive to not have the `,`.  And then just not do `String.split` at all in CfgAion.java.  For an example, see how CfgAion.java handles `--id=`

AlexandraRoatis:
@aion-kelvin I agree in the general case, but here I think we should opt for setting the port to the default `30303` value when given an incorrect value as input.

aion-kelvin:
> @aion-kelvin I agree in the general case, but here I think we should opt for setting the port to the default `30303` value when given an incorrect value as input.

I'm ok with that.  @beidouz could we implement it this way?  Also, make sure it doesn't write to the config file if the port is detected to be not valid.  It should also print an error/warning.

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
beidouz: this parameter takes two arguments, first one is <ip> second one is <port>, only <port> is getting overwritten in this case.

modAionImpl/src/org/aion/zero/impl/config/CfgAion.java
beidouz: will do, thanks!

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -231,6 +231,20 @@ public ReturnType call(final String[] args, Cfg cfg) {
 
             // 5. options that can be influenced by the -d and -n arguments
 
+            if (options.getPort() != null) {
+                int port_number = Integer.parseInt(options.getPort());
```
AlexandraRoatis: `parseInt` throws an exception when given a string that is not a number. This should be handled without crashing the kernel. For example `./aion.sh -p test` will show a NumberFormatException instead of properly notifying the user that an incorrect value has been given as parameter.

To solve this issue you can surround the statement with a `try-catch` and print the adequate message inside the catch, or set a boolean flag to true and check for the flag in the if statement below.

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -231,6 +231,20 @@ public ReturnType call(final String[] args, Cfg cfg) {
...
+                if (port_number < 0 || port_number > 0xFFFF) {
+                    System.out.println(
+                            "\nport out of range: "
+                                    + port_number
+                                    + ", switching to the current port configuration: \n");
+                } else {
+                    // update port in the config file
+                    cfg.toXML(new String[] {"--p2p=" + "," + options.getPort()}, configFile);
```
AlexandraRoatis: This line will modify the initial config file. You can verify that `config/mainnet/config.xml` will have the value of the port. It also ends up writing the id in the initial file, which is not the desired outcome. To avoid these side-effects you can set the port value inside the `cfg` object and set the `overwrite` flag to `true`. Writing to file will be triggered correctly below due to the `overwrite` flag.

Additionally, you can see in the tests (e.g. options `-p 12345`) that the config file is overwritten twice (not necessary).

modAionImpl/src/org/aion/zero/impl/config/CfgAion.java
AlexandraRoatis: still missing the correct formatting 
```java
if (cond) {
    cmd; 
}
```

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -231,6 +231,20 @@ public ReturnType call(final String[] args, Cfg cfg) {
 
             // 5. options that can be influenced by the -d and -n arguments
 
+            if (options.getPort() != null) {
+                int port_number = Integer.parseInt(options.getPort());
+                if (port_number < 0 || port_number > 0xFFFF) {
+                    System.out.println(
+                            "\nport out of range: "
+                                    + port_number
+                                    + ", switching to the current port configuration: \n");
```
AlexandraRoatis: please add here what port will be used

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -231,6 +231,20 @@ public ReturnType call(final String[] args, Cfg cfg) {
 
             // 5. options that can be influenced by the -d and -n arguments
 
+            if (options.getPort() != null) {
```
AlexandraRoatis: I think this option can work with `-c, --config` as well. The `if (options.getPort() != null) {...}` statement can be moved above the `if (options.getConfig() != null) {...}` statement. This will require adding more test cases as well.

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
```diff
@@ -603,6 +607,285 @@ public void testConfig_withMigration(String option) {
...
+            // with invalid parameter
+            parameters.add(
+                    new Object[] {
+                        new String[] {opPort, INVALID_PORT}, RUN, expPathOnError, expPortOnError
+                    });
+            parameters.add(
+                    new Object[] {
+                        new String[] {opPort, "-12345"}, RUN, expPathOnError, expPortOnError
+                    });
+            // with testing port number
```
AlexandraRoatis: we also need here a string value for the port that is not a number, e.g. `test`

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
```diff
@@ -861,6 +875,9 @@ TaskPriority getBreakingTaskPriority(Arguments options) {
             if (options.getDirectory() != null) {
                 skippedTasks.add("--datadir");
             }
+            if (options.getPort() != null) {
```
AlexandraRoatis: Please add some tests that use the port option in the `CliTest.testCheckArguments` as well.

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
```diff
@@ -603,6 +607,285 @@ public void testConfig_withMigration(String option) {
...
+        assertThat(cfg.getExecGenesisFile())
+                .isEqualTo(new File(expectedPath, "config" + File.separator + genesisFileName));
+        assertThat(cfg.getExecForkFile())
+                .isEqualTo(new File(expectedPath, "config" + File.separator + forkFileName));
+        assertThat(cfg.getDatabaseDir()).isEqualTo(new File(expectedPath, "database"));
+        assertThat(cfg.getLogDir()).isEqualTo(new File(expectedPath, "log"));
+        assertThat(cfg.getKeystoreDir()).isEqualTo(new File(expectedPath, "keystore"));
+        // test port is updated
+        assertThat(Integer.toString(cfg.getNet().getP2p().getPort())).isEqualTo(expectedPort);
+
```
AlexandraRoatis: This method should also check that the initial config file port is unchanged.

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
beidouz: :+1:

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
aion-kelvin: I've noticed a tangentially-related issue that I think we should fix in this review.  Linux can be configured to require root permissions for opening certain ports.  By default, ports <= 1024 are need root, but it could be configured to be anything.  Example:

```
sergiu@aion-XPS-8930:~/repos/aion2$ ./aion.sh -p 140
[...]
18-12-05 14:28:41.107 INFO  GEN  [main]: loaded block <num=24, root=2c0448b8... l=32>
18-12-05 14:28:41.171 ERROR P2P  [main]: tcp-server-socket-exception.
java.net.SocketException: Permission denied
        at java.base/sun.nio.ch.Net.bind0(Native Method)
        at java.base/sun.nio.ch.Net.bind(Net.java:461)
        at java.base/sun.nio.ch.Net.bind(Net.java:453)
        at java.base/sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:227)
        at java.base/sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:80)
        at org.aion.p2p.impl1.P2pMgr.run(P2pMgr.java:194)
        at org.aion.zero.impl.AionHub.initializeHub(AionHub.java:205)
        at org.aion.zero.impl.AionHub.<init>(AionHub.java:125)
        at org.aion.zero.impl.blockchain.AionImpl.<init>(AionImpl.java:77)
        at org.aion.zero.impl.blockchain.AionImpl.<init>(AionImpl.java:55)
        at org.aion.zero.impl.blockchain.AionImpl$Holder.<clinit>(AionImpl.java:68)
        at org.aion.zero.impl.blockchain.AionImpl.inst(AionImpl.java:72)
        at org.aion.zero.impl.blockchain.AionFactory.create(AionFactory.java:34)
        at org.aion.Aion.main(Aion.java:223)
18-12-05 14:28:41.175 INFO  GEN  [main]: <node-started endpoint=p2p://32b3fee1-a64d-4b05-baca-0f7101394b63@0.0.0.0:140>
18-12-05 14:28:41.178 INFO  CONS [main]: <sealing-disabled>
```

Then the kernel continues running.  I don't think the input validation of our code should check for this (there's other things we can't anticipate i.e. they specified a port that's already in use).  Instead, can we put a try/catch around the code that opens the port and print out an error if the port can't be opened?

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
beidouz: added

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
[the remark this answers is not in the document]
beidouz: :+1:

modAionImpl/src/org/aion/zero/impl/cli/Cli.java
beidouz: :+1:

Issue #734, opened 2018-11-28
API for VM-Kernel interaction


AlexandraRoatis:
Refactor the current code to extract a common interface layer to a new API for the VM repository in order to decouple the kernel implementation from the specific VM used by it.

The kernel will use the new VM-API jar and a specific implementation jar (no longer relying on git submodules).

FVM, AVM and the hybrid combination of the two would be options for the implementation jar. 

**Changes which must be made on the kernel side:**
- DB must support arbitrary-length keys and values (at the "repository" level)
- allow writing 0-values to DB -- don't want it to delete the data?
- Aion Transaction type field to define "deploy to AVM" as being distinct from "deploy to FVM"
- TransactionExecutor replaced by BulkExecutor

aionick:
The API should consist of two elements: data types (for which the implementation is provided in the API) and interfaces.

The following is a list of core aspects of the VM API. There may of course be oversights here and this list may grow, though I don't expect it to grow substantially and I also do not expect much of what is here to not make it into the actual API:

**Data Types**
- ResultCode: holds the 'exit status' of a transaction that was executed.
- ExecutionResult: the side-effects, output and energy consumption of an executed transaction.
- ExecutionContext: the pertinent world-state contextual details at the start of executing some transaction.
- Log: a smart contract event log fired off as a result of executing some transaction.
- Address: the public address of an account.
- ExternalTransaction: a transaction that is not spawned by some other transaction.
- InternalTransaction: a transaction that is spawned only as a consequence of executing some other transaction.
- Block: a block in the blockchain.
- LightweightFuture: a Java `Future` but much more lightweight since we only need the `get()` method. _This is of course an optional feature since it is provided by the JDK, but would prove more efficient and less cumbersome for our purposes_.

**Interfaces**
- VirtualMachine: the interface of an arbitrary virtual machine.
- KernelInterface: the interface of an arbitrary kernel implementation. The goal of this interface is not so that we can swap out kernels but so that a virtual machine has a line of communication back to the kernel.
- TransactionInterface: an interface over transaction types, particularly the two transaction types listed above.

aionick:
`TransactionResult` is now replacing `TransactionResult` (previous kernel version), `ExecutionResult`, `AbstractExecutionResult`, and `IExecutionResult`.

I'm moving onto `TransactionContext` now, which will drag with it a bunch of the other classes.

One question we have to answer soon is: Where should new contract addresses be generated?

[...]

PR #747, opened 2018-12-11
Remove dependency on aion.base from rlp, crypto and p2p

AlexandraRoatis:
## Description

This change is part of the VM API extraction #734  but can be reviewed separately. The objective is to have the kernel know about the VM, but not vice versa (as the dependencies are set now).

This intermediary step extracts the dependency on `modAionBase` for modules that were only using some utility functionality: crypto, rlp, and p2p.

**The initial classes are still left in the base module because a full refactoring is too complicated to track and review. They have been deprecated to indicate that this refactoring should be completed before the next release.**

The refactoring for the VM related modules will continue separately and will be completed as part of the API extraction.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

-

## Verification

### DIFF
--- a/modUtil/src/org/aion/util/hex/Hex.java
+++ b/modUtil/src/org/aion/util/hex/Hex.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2017-2018 Aion foundation.
+ *
+ *     This file is part of the aion network project.
+ *
+ *     The aion network project is free software: you can redistribute it
+ *     and/or modify it under the terms of the GNU General Public License
+ *     as published by the Free Software Foundation, either version 3 of
+ *     the License, or any later version.
+ *
+ *     The aion network project is distributed in the hope that it will
+ *     be useful, but WITHOUT ANY WARRANTY; without even the implied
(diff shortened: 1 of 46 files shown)


AlexandraRoatis:
OK. Tracking the full refactoring in #748.

modUtil/src/org/aion/util/hex/Hex.java
```diff
@@ -0,0 +1,139 @@
...
+ *     Aion foundation.
+ *     <ether.camp> team through the ethereumJ library.
+ *     Ether.Camp Inc. (US) team through Ethereum Harmony.
+ *     John Tromp through the Equihash solver.
+ *     Samuel Neves through the BLAKE2 implementation.
+ *     Zcash project team.
+ *     Bitcoinj team.
+ */
+
+package org.aion.util.hex;
```
aion-kelvin: minor suggestion: I think a better package name would be org.aion.util.conversions.  

With the JDK9 module-info stuff, packages are things that get explicitly exported, so I think we should make an effort to organize packages a bit better (i.e. avoid creating many of them, if it makes sense to have one big package that groups many related things together).

PR #756, opened 2018-12-18
Replace modGui directory with submodule 

aion-kelvin:
## Notice

## Description

- Replace modGui directory with submodule

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

- aion_gui directory is cloned when cloning aion using: `git clone --recursive https://github.com/aionnetwork/aion.git --branch aion_gui_submod  aion5`
- then, if I uncomment the last line in `gradle.properties` (to include GUI in aion build), then `./gradlew build` builds modGui
- Can run GUI using: `aion_gui/aion_gui.sh`
- `./gradlew pack` creates a .tar.bz2 that has the GUI.  Can use aion_gui.sh inside it to run the GUI.

## Verification


AlexandraRoatis:
I think we should take this opportunity to reverse the dependency relationship between the GUI and the kernel. By this I mean that instead of making the GUI a submodule of the kernel, I think the kernel should be a submodule of the GUI. 

Since the GUI is not a crytical or necessary component of the kernel, I don't think we should enforce updating it with every new feature. By keeping the GUI as a submodule there is an implict contract that it can at any time be enabled and will work correctly with any kernel. Whereas if the kernel is a submodule of the GUI we clearly state that the current version of the GUI works with the kernel at that specific commit. Also, it allows for better versioning of the GUI and a clear separation of developping kernel features and integrating the GUI with a new kernel release.

@AionJayT @jeff-aion @qoire please comment if you agree or disagree with the above change request.

aion-kelvin:
> Since the GUI is not a crytical or necessary component of the kernel, I don't think we should enforce updating it with every new feature.

I see your reasoning for this.  My feeling is that if we don't enforce this, what will happen is that the GUI will never be updated with any new kernel feature.  In fact, there have already been changes applied to the kernel that break the GUI.  One example is the multiple configurations we have for custom/mainnet/testnet -- GUI is hardcoded to mainnet and only works if you've run `./aion.sh` once before you run the GUI and I haven't heard of any plans or requests for changing this.

The GUI has been in a 'sort of supported' state for some time... given what you wrote, I think it is time to consider just removing the GUI altogether.

jeff-aion:
I would avoid the loaded meaning of a "submodule" and focus more on what constitutes a logical dependency.  I wouldn't want to tightly bind the relationship between these 2 projects in such a direct way.

Hopefully, the kernel doesn't logically depend on the GUI.  If the GUI depends on an actual kernel (or just a component thereof), then it should import this as an opaque dependency, maintaining an independent release lifecycle (I wouldn't want someone changing/localizing a button requiring an entire kernel release process).

After this change, a matter of updating/releasing the GUI becomes more of a packaging/distribution concern, no longer dictating code-level relationships.

AionJayT:
I think the ideal for the GUI project should be just a plug-in jar and the kernel doesn't depend on it. But it needs a lot of works until we totally decouple the module dependency. this PR just a half-way to archive the goal.

aion-kelvin:
Short version: It sounds like we all think that the kernel and GUI's dev and release cycles to be made independent and the concept that each GUI release works with some version of the kernel, but not vice versa.  I think that's the right way to go, but achieving that requires some additional work and I think it should be considered a separate issue instead of being lumped into this PR.  This PR's motivation is to get the GPL-dependent code out of the Aion repo so we can switch license.  This PR doesn't make the "decouple kernel/GUI" situation any worse, so if the licensing issue is a priority, we should push this PR so we can unblock it.

Long version:

I agree that kernel should not depend on GUI and we should be able to develop and release both of these things in a very decoupled way.  We're part way there: in gradle.properties you can disable modGui and the build process will totally ignore it (including your IDE).  This achieves the "don't think we should enforce updating GUI with every new feature" point that @AlexandraRoatis mentioned.  

Things preventing decoupled dev/release: 
1. GUI released tied to kernel release (what @jeff-aion mentioned)
2. Having a GUI submodule in kernel makes people think that at every commit in the kernel has a corresponding GUI commit that is compatible with it (what @AlexandraRoatis mentioned)

The intent of this PR is to get the GPL-dependent code (GUI) out of the repo of the rest of the kernel so that the kernel can switch license.  Making the GUI a submodule of kernel is the fastest way to make this happen because we can easily re-use the build logic already in the kernel, which has assumptions about the code and directory structure.

Making the GUI reference the kernel (either through git submodule, or Maven dependency, or something else), instead of the kernel referncing the GUI (how it is today), as @AlexandraRoatis suggested will improve problem 1 and fix problem 2.  It also creates additional work:

1. Gradle build logic needs to be changed because directory structure changes and right now the Gradle logic inside modGui can't work properly without the rest of the logic that's in the kernel
1. There's been talk of git submodule being not ideal, so should probably try to figure out a smarter way for doing referencing/depending on code
1. Need to figure out release process for kernel and GUI -- is kernel the kernel .tar.bz2 going to still include GUI?  If yes, then the problem of decoupling kernel and GUI releases is not fixed.  If no, then need to create additional instructions/logic for the GUI's .tar.bz2 package to be "plugged in" to the kernel's .tar.bz2 package.

aion-kelvin:
Ready to review again.  Bringing in spongycastle from Maven caused problems with one of aion_gui's other dependency, zxing because both of their jar names match the "core-X.Y.Z.jar" pattern, causing both of them to get the Java module name "core."  I modified aion_gui so the zxing jar gets renamed, which fixes that problem.

PR #761, opened 2018-12-20
Extracting an API for the Virtual Machine

AlexandraRoatis:
## Description

Refactoring the kernel to extract a VM API to be used for integration with both the FVM and AVM.

Fixes Issue #734.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing test suite

## Verification


modAion/src/org/aion/zero/types/AionTransaction.java
```diff
@@ -246,21 +258,23 @@ public void setTimeStamp(long timeStamp) {
...
     public byte[] getData() {
         if (!parsed) {
             rlpParse();
         }
         return data;
     }
 
-    public byte getType() {
+    @Override
+    public byte getTargetVM() {
```
AionJayT: just think we might have different transaction type implement. Is it okay to use it for the target VM?

modAion/test/org/aion/types/AionTransactionTest.java
```diff
@@ -121,8 +122,8 @@ public void testTransactionCost() {
     @Test
     public void testTransactionCost2() {
         byte[] nonce = DataWord.ONE.getData();
-        byte[] from = RandomUtils.nextBytes(Address.ADDRESS_LEN);
-        Address to = Address.EMPTY_ADDRESS();
+        byte[] from = RandomUtils.nextBytes(Address.SIZE);
```
AionJayT: should be AionAddress.SIZE?

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -548,12 +551,9 @@ public synchronized void compactState() {
...
     }
 
-    /**
-     * Processes a new block and potentially appends it to the blockchain, thereby changing the
-     * state of the world. Decoupled from wrapper function {@link #tryToConnect(AionBlock)} so we
-     * can feed timestamps manually
-     */
-    ImportResult tryToConnectInternal(final AionBlock block, long currTimeSeconds) {
+    // TEMPORARY: here to support the ConsensusTest
+    public Pair<ImportResult, AionBlockSummary> tryToConnectAndFetchSummary(
```
AionJayT: use @VisibleForTesting ?

modAionImpl/src/org/aion/zero/impl/StandaloneBlockchain.java
```diff
@@ -388,6 +391,12 @@ public synchronized ImportResult tryToConnect(final AionBlock block) {
         return result;
     }
 
+    // TEMPORARY: here to support the ConsensusTest
+    public synchronized Pair<ImportResult, AionBlockSummary> tryToConnectAndFetchSummary(
```
AionJayT: use     @VisibleForTesting?

modAionImpl/src/org/aion/zero/impl/valid/TXValidator.java
```diff
@@ -47,12 +47,13 @@
             Collections.synchronizedMap(new LRUMap<>(128 * 1024));
 
     public static boolean isValid(AionTransaction tx) {
-        Boolean valid = cache.get(ByteArrayWrapper.wrap(tx.getHash()));
+        ByteArrayWrapper transactionHash = (tx.getTransactionHash() == null) ? null : ByteArrayWrapper.wrap(tx.getTransactionHash());
+        Boolean valid = cache.get(transactionHash);
```
AionJayT: if transactionHash == null, this line might throw NPE

modDbImpl/src/org/aion/db/generic/DatabaseWithCache.java
```diff
@@ -425,69 +425,81 @@ public boolean isEmpty() {
...
-
+    public void putToBatch(byte[] k, byte[] v) {
+        AbstractDB.check(k);
+        AbstractDB.check(v);
         check();
 
-        for (Map.Entry<byte[], byte[]> entry : inputMap.entrySet()) {
-            ByteArrayWrapper key = ByteArrayWrapper.wrap(entry.getKey());
-            byte[] value = entry.getValue();
+        ByteArrayWrapper key = ByteArrayWrapper.wrap(k);
```
AionJayT: any benefit for using ByteArrayWrapper in this Cache class, can we use byte[] directly to avoid twice conversion?

modVM/build.gradle
```diff
@@ -5,6 +5,15 @@ dependencies {
     compile project(':modAion')
     compile project(':modMcf')
     compile project(':modCrypto')
+    compile project(':aion_vm_api')
+    compile project(':aion_fastvm')
+    compile files('../lib/org-aion-avm-core.jar')
+    compile files('../lib/org-aion-avm-rt.jar')
+    compile files('../lib/asm-6.2.1.jar')
```
AionJayT: can these binary download from maven repo instead of put in lib folder?

modPrecompiled/src/org/aion/precompiled/PrecompiledResultCode.java
```diff
@@ -0,0 +1,96 @@
...
+
+    STATIC_MODE_ERROR(8, ResultCategory.FAILED),
+
+    ABORT(11, ResultCategory.FAILED),
+
+    VM_REJECTED(-1, ResultCategory.FATAL),
+
+    VM_INTERNAL_ERROR(-2, ResultCategory.FATAL);
+
+    private enum ResultCategory {SUCCESS, REJECTED, FAILED, FATAL, REVERT }
```
aion-kelvin: Are these categories mutually exclusive?  If not, I don't think enum is a great way to model them

modPrecompiled/src/org/aion/precompiled/PrecompiledResultCode.java
```diff
@@ -0,0 +1,96 @@
+package org.aion.precompiled;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.aion.vm.api.interfaces.ResultCode;
+
+/**
+ * An enumeration representing the execution status of a transaction.
+ */
+public enum PrecompiledResultCode implements ResultCode {
```
aion-kelvin: How come FastVm, Avm, and Precompiled all have their own ResultCode class?  I worry that this will lead us down a path where different VMs give different result codes for the same result / a situation where the caller has to know about the underlying VM in order to interpret the result code.  Can we make them use the same set of result codes?

modVM/src/org/aion/vm/VirtualMachineFactory.java
```diff
@@ -0,0 +1,131 @@
+package org.aion.vm;
+
+import org.aion.avm.core.AvmImpl;
+import org.aion.avm.core.CommonAvmFactory;
+import org.aion.fastvm.FastVirtualMachine;
+import org.aion.vm.api.interfaces.KernelInterface;
+import org.aion.vm.api.interfaces.VirtualMachine;
+
+/**
+ * A Singleton factory class that is responsible for:
```
aion-kelvin: A few thoughts on this class...

Not a big fan of singleton pattern (for the reasons summarized in https://softwareengineering.stackexchange.com/a/40610) -- is it necessary to have that restriction?  I think it would be better to instantiate it in Aion.main and pass it to the places that need it.

Also, right now it can't be subclassed (due to `final` and the singleton pattern).  Since the vision is that people can implement their own VM if they like and use it in our code, that seems counter-intuitive.

I think 'Factory' is a bit of a misleading name since the class does work other than constructing virtual machines (it tracks their state, etc).  Would it be better to call it VirtualMachineManager or something like that?

aion-kelvin:
BTW, is there / will there be a PR for the changes in vm_api?  There's some changes in this PR that depend on it (i.e. I had some thoughts about Address, which is in vm_api, but AionAddress which implements it is in this PR) that I had some thoughts on.

modAion/src/org/aion/zero/types/AionTransaction.java
aionick: Yes, I was also thinking that `getTargetVM` may be too restrictive also. Currently transactions have a notion of a `type` and a `kind`, and this is really confusing and non-informative, so I would like to improve one of them. I'll try and think through some ideas, maybe `kind` should be the one to change

modPrecompiled/src/org/aion/precompiled/PrecompiledResultCode.java
aionick: Yes they are mutually exclusive. There is only one exception to this rule, where a failure that is caused by a `revert()` is considered both `FAILED` and `REVERT`, and that is only because revert causes a state change that is otherwise not caused by normal failures. In all other cases though, and going forward into the future, these should be mutually exclusive concepts.

modPrecompiled/src/org/aion/precompiled/PrecompiledResultCode.java
aionick: This is one of the things that is still in progress.
We are considering making everything except `FAILURE` the same and put into the API and allowing a VM to declare their own `FAILURE` types, since these are run-time failures, which may differ between VM's.
Not sure how we will handle this yet, but the issue is known

aionick:
The VM API changes were pushed straight to master and made into a first release, this is so that the AVM history can be maintained properly and meaningfully with respect to this artifact. All that was intended for the first release was: get this into a rough shape such that the kernel and its VMs can all co-exist. The next step is to figure out what belongs in the API and what doesn't, both in terms of interface functionality and also straight data-types that should instead be brought right into the API layer.
Any comments on that are welcome though

modAion/src/org/aion/zero/types/AionTransaction.java
```diff
@@ -208,6 +217,12 @@ public BigInteger getNonceBI() {
         return new BigInteger(1, getNonce());
     }
 
+    @Override
+    public byte[] getTimestamp() {
```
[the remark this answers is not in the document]
aionick: the `getTimeStamp()` method has been removed, so that we just have this one, `getTimestamp()`

modAionImpl/test/org/aion/zero/impl/cli/CliTest.java
```diff
@@ -403,6 +425,13 @@ private Object parametersWithConfig() {
             parameters.add(new Object[] {new String[] {op, "testnet"}, testnetConfig, expected});
         }
 
+        expected = AVM_TEST_BASE_PATH.getAbsolutePath();
+
+        for (String op : options) {
+            // avmtestnet as parameter
+            parameters.add(new Object[] {new String[] {op, "avmtestnet"}, avmtestnetConfig, expected});
```
arajasek: My code formatter reformats this.

aionick:
my bad on the force push

modAion/src/org/aion/zero/types/AionInternalTx.java
```diff
@@ -192,16 +199,16 @@ public void sign(ECKey key) throws ECKey.MissingPrivateKeyException {
 
     @Override
     public String toString() {
-        String to = (getTo() == null) ? "" : getTo().toString();
+        String to = (this.getDestinationAddress() == null) ? "" : this.getDestinationAddress().toString();
```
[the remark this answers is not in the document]
aionick: I removed the `this`

modDbImpl/src/org/aion/db/generic/DatabaseWithCache.java
aionick: @AlexandraRoatis would have better insight into this. I think it's to take advantage of the `equals` mostly?

modAionImpl/src/org/aion/zero/impl/valid/TXValidator.java
aionick: I believe the solution here should be that if the transactionHash is null, to compute `isValid0` on the transaction and return the result without added anything to the cache, otherwise, Kelvin is right, we could end up with a null key pointing to an answer that other null transactionHash transactions will inherit.

But this change is consensus-breaking, and we probably want to open a separate issue for this and do it at the same time as another fork.

For now, I will simply take care of the NPE.

aionick: I don't believe a null transactionHash is possible. The check probably isn't necessary at all. The `getTransactionHash()` method cannot return null. Agreed?

modVM/src/org/aion/vm/VirtualMachineFactory.java
aionick: I've changed the name to manager.
I do think the singleton/final pattern is appropriate for this class though, because we only want one authoritative manager. If a new vm implementation is added, that's no worry, you simply add it to the supported vm's in the manager (the manager itself is not a virtual machine, so subclassing is not an issue here).
However, the one exception to this is testing, where I agree we may want to mock the vm, so I've created a Provider class that provides the caller with the manager. This is where a test can set their mock manager, in the provider. This way the actual implementation doesn't have to inherit these testing concerns

AlexandraRoatis:
to be added in stages

Issue #758, opened 2018-12-20
API: estimateGas does not return correct values for contract creation txs.


arajasek:
> I'm running:
>
> - **Which Aion version?**: 0.3.2
> - **Which operating system?**: Linux
> - **Which distributor and version?**: Ubuntu 16.04
> - **How installed?**: from source
> - **Are you fully synchronized?**: yes
> - **Did you try to restart the node?**: yes

Contract creation transactions have a minimum gas value of 200,000. However, when `estimateGas` is called with a contract creation transaction (indicated by an empty `to` field), the value returned is sometimes below 200,000. 

As an example, the following web3 call currently returns an invalid estimate of 90,000.

`web3.eth.estimateGas({value:0,data:0x1}).then(console.log)`

arajasek:
The simplest workaround to this bug is to avoid using the `estimateGas` API call for contract creation transactions for now. The minimum gas value for such transactions is 200,000. I suggest using a value of 250,000 to ensure successful contract deployment.

arajasek:
Fixed in PRs #762 and #764 .

PR #762, opened 2018-12-21
Quickfix for estimating gas of contract creation txs

arajasek:
## Notice

**This PR is directly submitted to the master branch so we can rush-release a patch**

## Description

We have a bug in our API's `estimateGas` feature, as described in Issue #758.

This PR includes a quick fix for this, by using the maximum possible gas value when estimating gas for contract creation transactions.

The root cause of the issue also requires some fixes to other functionality, but that will be put into a later PR that goes into the next release.

## Type of change

Insert **x** into the following checkboxes to confirm (eg. [x]):
- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.
- [ ] Breaking change (a fix or feature that causes existing functionality to not work as expected).
- [ ] Requires documentation update.

## Testing

All tests pass, the example tx described in issue #758 returns a correct value now.

## Verification

Insert **x** into the following checkboxes to confirm (eg. [x]):
- [x] I have self-reviewed my own code and conformed to the style guidelines of this project.
- [x] New and existing tests pass locally with my changes.
- [ ] I have added tests for my fix or feature.
- [ ] I have made appropriate changes to the corresponding documentation.
- [x] My code generates no new warnings.
- [x] Any dependent changes have been made.

PR #764, opened 2019-01-03
API Server: Use different default gas values depending on tx type

arajasek:
## Notice

## Description

We currently do not differentiate between contract creation transactions and regular transactions. This is a problem, especially for gas-related issues because contract creation needs more gas. See issue #758 for an example.

This PR changes the default gas value supplied to a transaction depending on whether the tx is a contract creation call or not. It also does some code cleanup.

Fixes Issue #758  .

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

All tests pass, one new test added.

## Verification


modApiServer/src/org/aion/api/server/types/ArgTxCall.java
```diff
@@ -66,7 +66,7 @@ public ArgTxCall(
     }
 
     public static ArgTxCall fromJSON(
-            final JSONObject _jsonObj, NrgOracle oracle, long defaultNrgLimit) {
+            final JSONObject _jsonObj, long defaultNrgPrice, long txNrg, long contractCreateNrg) {
```
AlexandraRoatis: I also feel that these values should be constants instead of method parameters.

modApiServer/src/org/aion/api/server/ApiAion.java
```diff
@@ -769,13 +769,12 @@ protected long getRecommendedNrgPrice() {
         }
     }
 
-    // leak the oracle instance. NrgOracle is threadsafe, so safe to do this, but bad design
-    protected NrgOracle getNrgOracle() {
-        return NRG_ORACLE;
+    protected long getDefaultTxNrg() {
+        return 90_000L;
     }
```
AionJayT: What's the reason for deleting this method? It is be using for measuring reasonable price for the recently blocks.

modApiServer/src/org/aion/api/server/types/ArgTxCall.java
AionJayT: Agree!

arajasek: Done

arajasek:
> there are 4 methods call ArgTxCall.fromJSON(), It needs to be given different energy limit settings for avoiding the api call failed. e.g. the gas estimate should assign the max of the energy consume the kernel allowed instead of using the default.

This is fine, because the `estimateGas` case is handled seperately in the quickfix in PR #762, which is already in `master`

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -672,7 +672,10 @@ public RpcMsg eth_signTransaction(Object _params) {
             return new RpcMsg(null, RpcError.INVALID_PARAMS, "Invalid parameters");
         }
 
-        ArgTxCall txParams = ArgTxCall.fromJSON(_tx, getNrgOracle(), getDefaultNrgLimit());
+        ArgTxCall txParams =
```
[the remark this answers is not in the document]
arajasek: Fixed, sorry about that.

modMcf/src/org/aion/mcf/vm/Constants.java
```diff
@@ -33,6 +33,8 @@
 
     public static final int NRG_TX_CREATE = 200000;
 
+    public static final int NRG_TX_CREATE_DEFAULT = 350000;
```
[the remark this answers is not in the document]
arajasek: Done :)

PR #774, opened 2019-01-16
Step 1: kernel integration (use vm API)

aionick:
This is the first half of the work done on PR #761. The point of this separate PR is to try and get this batch of commits moved into `master-pre-merge`, since these have been identified as mostly simple re-naming and moving refactorings or the like. There is none of the core logic here.

Once we've established this branch is passing the CI and the other branch, rebased onto this (there were a couple inconsistencies to resolve), is also passing, this can probably be merged in right away if all approve (it seems so far we do) and the other PR can be updated to the rebased versions I've done on separate branches (for now).


modAionBase/src/org/aion/base/db/IRepositoryCache.java
```diff
@@ -78,5 +79,5 @@
      * @param key the key at which the date will be stored
      * @param value the data to be stored
      */
-    void addStorageRow(Address address, DW key, DW value);
+    void addStorageRow(Address address, ByteArrayWrapper key, ByteArrayWrapper value);
```
aion-kelvin: I noticed that as a result of this change, a lot of calls to it IRepositoryCache.addStorageRow has changed from:

`IRepositoryCache.addStorageRow(someAddress, new DataWord(someKey1), new DataWord(someKey2)`

to

`IRepositoryCache.addStorageRow(someAddress, new DataWord(someKey1).toWrapper(), new DataWord(someKey2).toWrapper()`

This happens all over the code and tests, i.e. StandaloneBlockchain.java:313-314. 

Would it make more sense to avoid that "add toWrapper() to everything" change by doing something like this:

- create an interface called ByteArrayWrappable with a method toWrapper()
- make IDataWord implement ByteArrayWrappable (and any other suitable class)
- either: (1) replace IRepositoryCache.addStorageRow ByteArrayWrapper arguments with ByteArrayWrappable and in each implementation, call ByteArrayWrappable.toWrapper() to get the ByteArrayWrapper
- or: (2) add a default method into IRepositoryCache: addStorageRow(Address, ByteArrayWrappable, ByteArrayWrappable) and all it does is call toWrapper() and pass it on to addStorageRow(Address address, ByteArrayWrapper key, ByteArrayWrapper value)

modAionBase/src/org/aion/base/type/ITransaction.java
```diff
@@ -1,48 +1,22 @@
 package org.aion.base.type;
 
 import java.math.BigInteger;
+import org.aion.vm.api.interfaces.TransactionInterface;
 
 /** @author jin */
-public interface ITransaction extends Cloneable {
-
-    byte[] getHash();
+public interface ITransaction extends Cloneable, TransactionInterface {
```
aion-kelvin: The ITransaction/TransactionInterface thing is a bit confusing, but as we talked about in-person, it's ok as like a intermediate step that will be eventually refactored.  Could you just either add a javadoc or TODO comment here or make a Github/Jira issue about this so we remember to revisit some time in the near future?

modAionBase/src/org/aion/base/type/AionAddress.java
```diff
@@ -12,51 +12,49 @@
...
-    private static final Address emptyAddr = Address.wrap(new byte[0]);
+public final class AionAddress implements org.aion.vm.api.interfaces.Address, Comparable<AionAddress>, Bytesable<AionAddress>, Cloneable {
+    private static final AionAddress zeroAddr = AionAddress.wrap(new byte[SIZE]);
+    private static final AionAddress emptyAddr = AionAddress.wrap(new byte[0]);
 
     private byte[] address;
     private int hashCode = 0;
 
-    public Address(final byte[] in) {
+    public AionAddress(final byte[] in) {
```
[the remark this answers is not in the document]
AlexandraRoatis: This is done in a later commit and will be updated in the second PR

modAionBase/src/org/aion/base/util/ByteArrayWrapper.java
```diff
@@ -61,4 +61,32 @@ public ByteArrayWrapper fromBytes(byte[] bs) {
...
+     * @return {@code true} if every byte in the array has the value 0, {@code false} otherwise
+     */
+    public boolean isZero() {
+        int length = data.length;
+        for (int i = 0; i < length; i++) {
+            if (data[length - 1 - i] != 0) {
+                return false;
+            }
+        }
+        return true;
```
AionJayT: this method might need to benchmark to see this implementation is faster than pure byte compares by order.

modAionBase/src/org/aion/base/db/IRepositoryCache.java
AlexandraRoatis: We could make this comment an issue that can be verified at the end, once all stages of the PR are added.

modMcf/src/org/aion/mcf/vm/types/DataWord.java
```diff
@@ -54,7 +54,31 @@ public DataWord(byte[] data) {
...
+     *
+     * @param number The {@link BigInteger} whose byte array representation is to be possibly
+     *     truncated.
+     * @return The re-formatted {@link BigInteger#toByteArray()} representation as here specified.
+     */
+    private static byte[] removeLargeBigIntegerLeadingZeroByte(BigInteger number) {
+        byte[] bytes = number.toByteArray();
+        return ((bytes.length == (DataWord.BYTES + 1)) && (bytes[0] == 0x0))
+                ? Arrays.copyOfRange(bytes, 1, bytes.length)
+                : bytes;
```
AionJayT: how to handle the case when the Bignumber input more than 16 bytes?

modAionBase/src/org/aion/base/type/AionAddress.java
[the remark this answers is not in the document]
AlexandraRoatis: existing commit for fixing the issue: https://github.com/aionnetwork/aion/pull/761/commits/ba3431d16e90f0366bac0744a63f7bf54fe8628b

modMcf/src/org/aion/mcf/vm/types/DataWord.java
aionick: This byte array gets passed into the constructor, which performs the check

PR #778, opened 2019-01-17
Step 2: kernel integration (migrate functionality to fvm)

aionick:
Step 2 in the kernel integration.

This PR mostly moves classes from the kernel into the fast vm.

A consensus test has been added to check some basic hard-coded hash values that we must also produce to not break consensus. This test will be updated as we go on.

PR #779, opened 2019-01-17
Step 3: kernel integration

aionick:
Step 3 in the integration.

This PR is mostly concerned with setting up the `BulkExecutor` class, which is the class that handles virtual-machine processing requests in the kernel. This class will eventually be made to receive a list of transactions and then to delegate them to their appropriate virtual machine.

Some moving around of tests was necessary, as well as some movement of logic from the kernel side over to the fvm side.

What we have found out about some of the fvm tests in the process: they know too intimately about the objects they touch; some tests will need considerable re-writing.

**_This PR is currently merging into step 2 of the integration (PR #778). This is only to produce a meaningful diff. Once that PR is merged in to master-pre-merge this PR will be updated to merge onto that branch as well._**


modPrecompiled/test/org/aion/precompiled/contracts/TXHashContractTest.java
```diff
@@ -63,6 +64,13 @@ public void setUp() throws IOException {
                         0L,
                         0L,
                         null);
+=======
+        ExecutionContext ctx = new ExecutionContext(null, txHash,
+            ContractFactory.getTxHashContractAddress(), null, null, null,
+            0L, null, null, 0, 0, 0, null,
+            2000001L, 0L, 0L,
+            null);
+>>>>>>> 795ef66... Fastvm implements VirtualMachine
```
AionJayT: the merge conflict should fix

modAionImpl/test/org/aion/zero/impl/vm/Benchmark.java
```diff
@@ -0,0 +1,308 @@
+/*
+ * Copyright (c) 2017-2018 Aion foundation.
```
aion-kelvin: unnecessary header

modAionImpl/src/org/aion/zero/impl/blockchain/AionImpl.java
```diff
@@ -159,20 +161,31 @@ public AionTxReceipt callConstant(AionTransaction tx, IAionBlock block) {
...
         } finally {
             repository.rollback();
         }
     }
 
+    /**
+     * There is no post-execution work to do for any calls in this class.
+     */
+    private PostExecutionWork getPostExecutionWork() {
+        return (k, s, t, b) -> {
```
aion-kelvin: Can fix this later, but I think it would make far more sense to instantiate the PostExecutionWork once at constructor (or static) time and return that instance all the time, instead of instantiating it every time this getter is called.

PR #791, opened 2019-01-21
Step 8: kernel integration (CLI for avmtestnet + config to disable AVM use)

AlexandraRoatis:
Step 6 in the integration.

This PR is concerned with adding the necessary configuration for:
1. running the kernel on the `avmtestnet`;
2. disabling AVM functionality on the production networks.

**Requires a rebase after Step 5 is merged.**


modDbImpl/src/org/aion/db/impl/leveldb/LevelDB.java
```diff
@@ -334,73 +334,30 @@ public boolean hasNext() {
...
-            // bulk atomic update
-            db.write(batch);
-        } catch (DBException e) {
-            LOG.error(
-                    "Unable to execute batch put/update operation on " + this.toString() + ".", e);
-        } catch (IOException e) {
-            LOG.error("Unable to close WriteBatch object in " + this.toString() + ".", e);
+    public void putToBatchInternal(byte[] key, byte[] value) {
+        if (batch == null) {
+            batch = db.createWriteBatch();
```
AionJayT: does leveldb createWriteBatch guarantee it always success?

modDbImpl/src/org/aion/db/impl/leveldb/LevelDB.java
```diff
@@ -423,11 +380,31 @@ public void commitBatch() {
...
     @Override
-    public void deleteBatch(Collection<byte[]> keys) {
-        check(keys);
+    public void putBatchInternal(Map<byte[], byte[]> input) {
+        // try-with-resources will automatically close the batch object
+        try (WriteBatch batch = db.createWriteBatch()) {
+            // add put and delete operations to batch
+            for (Map.Entry<byte[], byte[]> e : input.entrySet()) {
+                byte[] key = e.getKey();
+                byte[] value = e.getValue();
```
AionJayT: can we just assign the return value into the following method?

AionJayT: to avoid the local variable creating increase the heap size and the GC

AionJayT: just not sure does 
                byte[] key = e.getKey();
                byte[] value = e.getValue();
                batch.put(key, value);
has any overhead during the runtime?
does this has same memory uausge and performance as just using  batch.put(e.getKey(), e.getValue);

modAionBase/src/org/aion/base/vm/VirtualMachineSpecs.java
```diff
@@ -0,0 +1,7 @@
+package org.aion.base.vm;
+
+public final class VirtualMachineSpecs {
+
+    public static final byte AVM_VM_CODE = 0xf;
+    public static final byte FVM_DEFAULT_TX_TYPE = 0x01;
```
aionick: `0x0f` is used when creating a new Avm contract so maybe `AVM_CREATE_CODE` is a better name

AlexandraRoatis: updated both the value and the name

modAionImpl/src/org/aion/zero/impl/config/Network.java
```diff
@@ -11,6 +11,7 @@
     MAINNET("mainnet", 256),
     CONQUEST("conquest", 128),
     MASTERY("mastery", 32),
+    AVMTESTNET("avmtestnet", 31), // temporary chainid
```
aion-kelvin: is it still tempoary?

AlexandraRoatis: yeah, changing the chain id is something that may still happen

PR #798, opened 2019-01-29
Step 7: kernel integration (repo supports arbitrary sized key values)

aionick:
The purpose of this PR is to allow for the repository to accept arbitrarily sized key-value pairings.

### DIFF
--- a/modAionImpl/test/org/aion/db/AionContractDetailsTest.java
+++ b/modAionImpl/test/org/aion/db/AionContractDetailsTest.java
@@ -11,6 +11,7 @@ import org.aion.base.db.IContractDetails;
 import org.aion.base.db.IPruneConfig;
 import org.aion.base.db.IRepositoryConfig;
 import org.aion.base.type.AionAddress;
+import org.aion.base.util.ByteArrayWrapper;
 import org.aion.base.util.ByteUtil;
 import org.aion.db.impl.DBVendor;
 import org.aion.db.impl.DatabaseFactory;
@@ -267,7 +268,7 @@ public class AionContractDetailsTest {
             DataWord value = new DataWord(RandomUtils.nextBytes(16));
 
             elements.put(key, value);
-            original.put(key.toWrapper(), value.toWrapper());
+            original.put(key.toWrapper(), wrapValueForPut(value));
         }
 
         original.syncStorage();
@@ -283,7 +284,7 @@ public class AionContractDetailsTest {
         assertEquals(ByteUtil.toHexString(code), ByteUtil.toHexString(deserialized.getCode()));
 
         for (DataWord key : elements.keySet()) {
-            assertEquals(elements.get(key).toWrapper(), deserialized.get(key.toWrapper()));
+            assertEquals(elements.get(key).toWrapper(), wrapValueFromGet(deserialized.get(key.toWrapper())));
         }
 
         DataWord deletedKey = elements.keySet().iterator().next();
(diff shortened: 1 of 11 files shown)


modMcf/src/org/aion/mcf/db/ContractDetailsCacheImpl.java
```diff
@@ -49,7 +50,14 @@ public static ContractDetailsCacheImpl copy(ContractDetailsCacheImpl cache) {
      */
     @Override
     public void put(ByteArrayWrapper key, ByteArrayWrapper value) {
-        storage.put(key, value);
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(value);
+
+        if (value.isZero()) {
+            storage.remove(key);
```
AlexandraRoatis: This was part of the put/delete separation that was delegated to a later commit. I worry it may cause consensus issues.

modAionImpl/test/org/aion/db/AionContractDetailsTest.java
```diff
@@ -336,7 +337,15 @@ public void testExternalStorageTransition() {
...
+            assertEquals(elements.get(key).toWrapper(), wrapValueFromGet(deserialized.get(key.toWrapper())));
         }
     }
+
+    private static ByteArrayWrapper wrapValueForPut(DataWord value) {
+        return (value.isZero()) ? new ByteArrayWrapper(value.getData()) : new ByteArrayWrapper(value.getNoLeadZeroesData());
+    }
+
+    private static ByteArrayWrapper wrapValueFromGet(ByteArrayWrapper value) {
+        return new ByteArrayWrapper(new DataWord(value.getData()).getData());
```
AlexandraRoatis: In the final clean-up stage, the `DataWord` should be completely removed from the kernel. Only FVM should know about it.

modPrecompiled/test/org/aion/precompiled/contracts/DummyRepo.java
```diff
@@ -122,7 +122,7 @@ public ByteArrayWrapper getStorageValue(Address addr, ByteArrayWrapper key) {
         Map<String, byte[]> map = storage.get(addr);
         if (map != null && map.containsKey(key.toString())) {
             byte[] res = map.get(key.toString());
-            if (res.length == DataWord.BYTES) {
+            if (res.length <= DataWord.BYTES) {
```
AlexandraRoatis: Where do we have instances that do not match the specified length?

modMcf/src/org/aion/mcf/db/ContractDetailsCacheImpl.java
aionick: I thought the separation pulled these two out into their own methods? What should be done here then?

modAionImpl/test/org/aion/db/AionContractDetailsTest.java
aionick: Agreed. It exists here simply because this is a test and we haven't gotten that far into the integration, but soon the kernel will lose visibility on `DataWord` and `DoubleDataWord`

modPrecompiled/test/org/aion/precompiled/contracts/DummyRepo.java
aionick: It was previous the repository's duty to take a `DataWord` and truncate it's leading zeroes, then insert it. Now the repository does nothing and this truncating burden is on the caller. Likewise, the caller now also inherits the burden of having to re-pad this data, and that's what's being done here. This catches any truncated data words and re-pads them

modMcf/src/org/aion/mcf/db/ContractDetailsCacheImpl.java
AlexandraRoatis: I think for now we can leave this method unchanged. The update for contract details will be added as a separate PR.

aionick: will do

PR #801, opened 2019-01-30
Move ZMQ native libs and native loading logic into 3rdParty/libnzmq

aion-kelvin:
## Notice

## Description

- Move ZMQ native libs and native loading logic into 3rdParty/libnzmq
  - put the .so files in libnzmq (when it builds, its .jar will have the .so files)
  - libnzmq seemed to have some native library loading code ... I don't think it was used and wasn't implemented correctly though, so I replaced it with a simpler implementation
  - update Gradle
  - removed the library loading code from rest of kernel
- removed some modules' Gradle dependency on aion_api (they weren't actually needed)
- aion_api needs an update for this to work -- https://github.com/aionnetwork/aion_api/pull/57

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

- Manual testing only
  - compiled and ran kernel with Api Server and connected Desktop Wallet to it 
  - `./gradlew pack` succeeds 

## Verification


aion-kelvin:
> this change looks like only support the linux system.  how about the supporting for the other systems? Because the libnzmq can load the native lib by the OS system before the change.

That is true, this will only work out-of-the-box on Linux.  I am considering adding support for the Mac and Windows in a later PR, but I think for this PR, this is ok.  Because before this change, this library didn't load the native library at all.  It depended on some other part of the kernel to call `System.load`.  

Also, this library can still be made to work on other OS's -- JVM needs the ZMQ native lib loaded, but it doesn't care what loads it.  So, for example-- DesktopWallet depends needs to use this lib, but DesktopWallet itself (or some library it uses) can call `System.load("myMacLibrary.dylib")` before using the ZMQ class in here.

Today, in kernel, modApiServer and modAionImpl call System.load on the ZMQ libs.  I'm trying to move those into 3rdParty/libnzmq.  Since kernel doesn't support Mac and Windows, I just put in Linux support for now.

> The desktop wallet seems to reply with this library. Make sure they are not using the lib in the other platform.

So, the changes in this PR, plus some changes I have planned for aion_api may need some changes on the Desktop Wallet side.  My plan is once aion_api is separated from kernel (that will be my next PR), I will check Desktop Wallet to see if it needs changes.  Until then, Desktop Wallet should continue to use the aion_api .jar (and other kernel jars).  I talked with the Centrys guys and they said they don't plan to have to change those jars in the next month or two, so I think that is ok.

Let me know what you think!

aion-kelvin:
Also: there's lots of merges into aion/master-pre-merge and aion_api/master.  These keep causing build errors because the aion_api submodule pointer changes and then git doesn't know how to auto-resolve it.  It's really easy to fix, but if you see the build check failed here, that's probably why.  It is still ok for reviewing though because I'm pretty sure the stuff being merged in does not touch this (except for the aion_api pointer)

AionJayT:
continue to discuss. Will it affect the other projects using Java API cause it loses the cross-platform support? I am okay with the kernel part. only concern is the desktop wallet or other projects support.

AionJayT:
discussed with @aion-kelvin , the client api will merge first then put the cross-platform support in the following PR,

aion-kelvin:
@AionJayT -- heard back from Centrys; they said:

> as long as the communication protocol between the java api and the kernel is not affected, we can live with an older version of the api.

So, I think we are good from their side!

PR #802, opened 2019-01-30
Step 9: kernel integration (move to canonical vm_api build)

aionick:
This moves us to the last vm_api commit so that we are compatible with the AVM.

This final vm_api commit just makes some name changes (to make merging easier in the AVM in certain cases, and because the names are often better anyway).

There is also one new method introduced, `getContractAddress()` in the `KernelInterface`.

This PR has corresponding changes in:
- [aion_api](https://github.com/aionnetwork/aion_api/pull/58)
- [aion_fastvm](https://github.com/aionnetwork/aion_fastvm/pull/50)

### DIFF
--- a/modAion/src/org/aion/zero/types/AionTransaction.java
+++ b/modAion/src/org/aion/zero/types/AionTransaction.java
@@ -251,7 +251,7 @@ public class AionTransaction extends AbstractTransaction {
     }
 
     @Override
-    public byte getTransactionType() {
+    public byte getTargetVM() {
         if (!parsed) {
             rlpParse();
         }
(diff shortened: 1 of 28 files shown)


modAion/src/org/aion/zero/types/AionTransaction.java
```diff
@@ -251,7 +251,7 @@ public Address getDestinationAddress() {
     }
 
     @Override
-    public byte getTransactionType() {
+    public byte getTargetVM() {
```
AlexandraRoatis: Can we reverse this renaming? After some discussion, it was concluded that it is too restrictive and we should keep the option of multiple transaction types.

aionick: We will reverse this on our next vm api iteration. Just because we have already tagged the build and made a release in the vm_api repository so that the AVM could have a reliable version history, so it's quite a bit of work to undo it right now

AlexandraRoatis: ok

PR #803, opened 2019-01-30
Step 10: kernel interface (Avm transactions supported)

aionick:
This PR lets the `BulkExecutor` invoke the Aion virtual machine.

Some tests have been added and other minor fixes.

There are corresponding submodule changes at:
- [aion_fastvm](https://github.com/aionnetwork/aion_fastvm/pull/51)

_Still have some testing & verification to do on this build; do not merge yet_


modAionImpl/src/org/aion/zero/impl/StandaloneBlockchain.java
```diff
@@ -364,7 +371,7 @@ public void assertEqualTotalDifficulty() {
 
     @Override
     public synchronized ImportResult tryToConnect(final AionBlock block) {
-        ImportResult result = tryToConnectInternal(block, System.currentTimeMillis() / 1000);
+        ImportResult result = tryToConnectInternalAndSkipTimeCheck(block, System.currentTimeMillis() / 1000);
```
AlexandraRoatis: I don't think we should make this change. For testing, we can create blocks that would pass the time check using `createNewBlockInternal`.

modAion/src/org/aion/zero/types/AionTransaction.java
```diff
@@ -89,6 +89,23 @@ public AionTransaction(
...
+    // constructor for explicitly setting a transaction type.
+    public AionTransaction(
+        byte[] nonce,
+        Address from,
+        Address to,
+        byte[] value,
+        byte[] data,
+        long nrg,
+        long nrgPrice,
+        byte txType) {
```
AlexandraRoatis: formatting

modAionImpl/src/org/aion/zero/impl/StandaloneBlockchain.java
aionick: Thanks, didn't know about this option, I will make the change

AionJayT:
LGTM

modVM/src/org/aion/vm/ExecutionBatch.java
```diff
@@ -127,10 +127,7 @@ private KernelTransactionContext constructTransactionContext(
         long energyLimit = transaction.nrgLimit();
         long nrg = transaction.nrgLimit() - transaction.transactionCost(block.getNumber());
         DataWord callValue = new DataWord(ArrayUtils.nullToEmpty(transaction.getValue()));
-        byte[] callData =
-                transaction.isContractCreationTransaction()
-                        ? ByteUtil.EMPTY_BYTE_ARRAY
-                        : ArrayUtils.nullToEmpty(transaction.getData());
+        byte[] callData = ArrayUtils.nullToEmpty(transaction.getData());
```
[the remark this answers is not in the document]
aionick: This is because of our close coupling, we didn't realize that this conditional was an fvm implementation detail and not a high level concept. 

This batch sends the contexts off to the VM's, and the AVM relies on this field being set as the actual data. The FVM, however, does not rely on this, and it made this conditional choice to anticipate its later serialization concerns, which the AVM does not have.

So at this level we should be passing the actual data. In the contexts that the FVM uses, however, we need to preserve this logic for the `toBytes()`

modVM/src/org/aion/vm/KernelTransactionContext.java
```diff
@@ -160,7 +169,7 @@ public Address getDestinationAddress() {
     @Override
     public Address getContractAddress() {
         byte[] rawBytes = this.transaction.getContractAddress().toBytes();
-        rawBytes[0] = 0x0f;
+        rawBytes[0] = NodeEnvironment.CONTRACT_PREFIX;
```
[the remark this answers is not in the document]
aionick: These are in fact different constants, it just happens that their current values are the same. The `AVM_CREATE_CODE` is the transaction type that the executor uses to determine which VM to use. But this is the contract prefix, the expected first byte of any AVM contract address, which is used for calls and for determining whether a VM is performing a cross-vm call. 

This might also be represented in a different way in the future, so good to not confuse them.

modApiServer/src/org/aion/api/server/types/ArgTxCall.java
```diff
@@ -52,6 +75,7 @@ public static ArgTxCall fromJSON(final JSONObject _jsonObj, long defaultNrgPrice
                     AionAddress.wrap(ByteUtil.hexStringToBytes(_jsonObj.optString("from", "")));
             Address to = AionAddress.wrap(ByteUtil.hexStringToBytes(_jsonObj.optString("to", "")));
             byte[] data = ByteUtil.hexStringToBytes(_jsonObj.optString("data", ""));
+            byte type = ByteUtil.hexStringToBytes(_jsonObj.optString("type", "0x1"))[0];
```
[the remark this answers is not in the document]
aionick: This line is just saying that, if no transaction type is specified in the RPC call, the default value to be used will be `0x1`. If someone explicitly sets the type to `0x0` then that's the type they get

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -1170,19 +1171,18 @@ private AionBlockSummary applyBlock(IAionBlock block) {
...
+     * <p>This "work" is specific to the {@link AionBlockchainImpl#applyBlock(IAionBlock)} method.
      */
     private static PostExecutionWork getPostExecutionWorkForApplyBlock() {
         return (topRepository,
-            childRepository,
-            transactionSummary,
-            transaction,
-            blockEnergyLeft) -> {
-
+                childRepository,
```
aion-kelvin: since the lambda function doesn't need to change at each call of this method, can do something like this to instantiate it once only:

```java
    private static PostExecutionWork PEW = (topRepository, childRepository, transactionSummary, transaction, blockEnergyRemaining) -> {
        if (!transactionSummary.isRejected()) {
            childRepository.flush();

            AionTxReceipt receipt = transactionSummary.getReceipt();
            receipt.setPostTxState(topRepository.getRoot());
            receipt.setTransaction(transaction);

            return receipt.getEnergyUsed();
        } else {
            return 0;
        }
    };

    private static PostExecutionWork getPostExecutionWorkForGeneratePreBlock2 () {
        return PEW;
    }
```

modAionImpl/test/org/aion/zero/impl/vm/StatefulnessTest.java
```diff
@@ -0,0 +1,281 @@
...
+import org.apache.commons.lang3.RandomUtils;
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Ignore;
+import org.junit.Test;
+
+// These tests are ignored for now because in order for them to pass we need the clock drift buffer
```
aion-kelvin: old comment?

modVM/src/org/aion/vm/KernelInterfaceForAVM.java
```diff
@@ -0,0 +1,190 @@
...
+
+        if (repositoryCache == null) {
+            throw new NullPointerException("Cannot set null repositoryCache!");
+        }
+        this.repositoryCache = repositoryCache;
+        this.allowNonceIncrement = allowNonceIncrement;
+        this.isLocalCall = isLocalCall;
+    }
+
+    // These 4 methods are temporary. Really any of this type of functionality should be moved out
```
aion-kelvin: is this still correct?

modVM/src/org/aion/vm/VmFactoryImplementation.java
```diff
@@ -0,0 +1,108 @@
...
+import org.aion.avm.core.AvmImpl;
+import org.aion.avm.core.CommonAvmFactory;
+import org.aion.fastvm.FastVirtualMachine;
+import org.aion.vm.api.interfaces.KernelInterface;
+import org.aion.vm.api.interfaces.VirtualMachine;
+
+/**
+ * A Singleton factory class that is responsible for:
+ *
+ * <p>- Initializing the state of all the supported virtual machines. - Returning any requested
```
aion-kelvin: weird formatting

modVM/src/org/aion/vm/VirtualMachineManager.java
```diff
@@ -0,0 +1,64 @@
...
+ * <p>The order of calls to a VirtualMachineManager must be done in the following order:
+ *
+ * <p>1. initializeAllVirtualMachines 2. getVirtualMachineInstance 3. shutdownAllVirtualMachines
+ *
+ * <p>where the second call can be repeated multiple times before a shutdown.
+ *
+ * <p>This ordering is to ensure that when getVirtualMachineInstance is called, the {@link
+ * VirtualMachine} it returns is initialized and ready to use.
+ */
+public interface VirtualMachineManager {
```
aion-kelvin: IMO, this factory isn't really needed, we can just use its implementation directly (it also depends on an inner class of its implementation VmFactoryImplementation.VM which prevents anyone from using this interface without also using that implementation)

Could you also use the word "Factory" instead of "Manager" in the name (of this interface and/or its impl)

config/avmtestnet/config.xml
```diff
@@ -60,7 +62,7 @@
         <!--Sets the physical location on disk where data will be stored.-->
         <path>database</path>
         <!--Boolean value. Enable/disable database integrity check run at startup.-->
-        <check_integrity>true</check_integrity>
+        <check_integrity>false</check_integrity>
```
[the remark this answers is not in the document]
AlexandraRoatis: This is not really necessary on any of the test networks.

modVM/src/org/aion/vm/VirtualMachineManager.java
aionick: I'm going to post an issue about this, it may be a large enough refactoring to warrent more consideration, and I don't want to be changing too much logic at this point.

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
aionick: I'm going to also move this into the next set of refactorings, because I don't want to alter logic at this point, there's pressure to get these changes in

PR #809, opened 2019-02-05
Step 11: kernel integration (put / delete split in contract details)

AlexandraRoatis:
## Description

Last important integration step. Separating the put and delete in contract details to later allow the AVM to put bute arrays that are all 0 values.

Requires changes in [aion_fastvm](https://github.com/aionnetwork/aion_fastvm/pull/52).

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

- existing test suite
- an additional test to ensure that the `ContractDetailsCacheImpl` behaves correctly on put and delete

## Verification

### DIFF
--- a/modAionBase/src/org/aion/base/db/IContractDetails.java
+++ b/modAionBase/src/org/aion/base/db/IContractDetails.java
@@ -9,14 +9,19 @@ import org.aion.vm.api.interfaces.Address;
 public interface IContractDetails {
 
     /**
-     * Inserts key and value as a key-value pair. If the underlying byte array of value consists
-     * only of zero bytes then ay existing key-value pair that has the same key as key will be
-     * deleted.
+     * Inserts a key-value pair containing the given key and the given value.
      *
-     * @param key The key.
-     * @param value The value.
+     * @param key the key to be inserted
+     * @param value the value to be inserted
      */
     void put(ByteArrayWrapper key, ByteArrayWrapper value);
+
+    /**
+     * Deletes any key-value pair that matches the given key.
+     *
+     * @param key the key to be deleted
+     */
+    void delete(ByteArrayWrapper key);
 
     /**
      * Returns the value associated with key.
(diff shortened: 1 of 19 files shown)


modAionImpl/src/org/aion/zero/impl/db/AionRepositoryDummy.java
```diff
@@ -139,24 +139,33 @@ public BigInteger getBalance(Address addr) {
...
+                            + key.toString()
+                            + " which is not a valid stored value for the FVM. ");
         }
 
-        details.put(key, value);
-        detailsDB.put(ByteArrayWrapper.wrap(addr.toBytes()), details);
+        return value;
     }
 
+    // never used
```
aion-kelvin: delete?

modAionBase/src/org/aion/base/db/IContractDetails.java
```diff
@@ -113,27 +117,19 @@
...
-     * @param storageValues The values.
-     */
-    void setStorage(List<ByteArrayWrapper> storageKeys, List<ByteArrayWrapper> storageValues);
-
     /**
      * Sets the storage to contain the specified key-value mappings.
      *
-     * @param storage The specified mappings.
+     * @param storage the specified mappings
+     * @apiNote Used for testing.
```
aion-kelvin: @VisibleForTesting?

AlexandraRoatis: I marked the fact that it was only used for testing so that maybe it gets removed from the interface during the refactoring. Then the tag can be added to the implementations.

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryDummy.java
AlexandraRoatis: We may want to remove the full class, cause there are multiple classes used for repository test functionality. But this feels like a separate task.

AlexandraRoatis: If this is the test class that ends up remaining after refactoring, the functionality may be reenabled.

PR #813, opened 2019-02-07
Cleaning up some trie functionality

AlexandraRoatis:
## Description

1. Commented out some unused functionality.
2. Cleaned up the object construction process by adding a constructor using a cache.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing test suite + `--redo-import` on mastery

## Verification


modMcf/src/org/aion/mcf/trie/TrieImpl.java
```diff
@@ -876,16 +871,10 @@ private void traceTrie(byte[] stateRoot, ScanAction action) {
 
     public boolean validate() {
         synchronized (cache) {
```
AionJayT: inside the scanTreeLoop already synchronized the cache, Can we remove the synchronized in this method?

AlexandraRoatis: This `validate` method is actually not used either. Should I completely remove it?

AionJayT: if we confirmed the method is thread-safe. we should remove it.
The core components will affect some performance of the kernel.

AlexandraRoatis: Remove the extra synchronization and commented out the method. It can be re-enabled and used correctly if ever needed.

PR #814, opened 2019-02-07
Sync State 1: Messages for trie data requests/responses

AlexandraRoatis:
## Description

As part of a fast sync implementation, the nodes need the ability to communicate trie data to each other.
This is the first step in that direction containing the message transmitted between nodes to communicate this data.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [x] Unit test.

## Testing

- existing tests + new test for the added functionality

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/sync/msg/RequestTrieData.java
+++ b/modAionImpl/src/org/aion/zero/impl/sync/msg/RequestTrieData.java
@@ -1,0 +1,134 @@
+package org.aion.zero.impl.sync.msg;
+
+import java.math.BigInteger;
+import java.util.Objects;
+import org.aion.p2p.Ctrl;
+import org.aion.p2p.Msg;
+import org.aion.p2p.Ver;
+import org.aion.rlp.RLP;
+import org.aion.rlp.RLPList;
+import org.aion.zero.impl.sync.Act;
+import org.aion.zero.impl.sync.DatabaseType;
+
+/**
+ * Request message for a trie node from a specific blockchain database.
+ *
+ * @author Alexandra Roatis
+ */
+public final class RequestTrieData extends Msg {
+    private final DatabaseType dbType;
+    private final byte[] nodeKey;
+    private final int limit;
+
+    /**
+     * Constructor for trie node requests with specified limit.
+     *
+     * @param nodeKey the key of the requested trie node
+     * @param dbType the blockchain database in which the key should be found
+     * @param limit the maximum number of key-value pairs to be retrieved by the search inside the
+     *     trie for referenced nodes
+     * @throws NullPointerException if either of the given parameters is {@code null}
+     * @throws IllegalArgumentException if the given limit is negative
+     */
+    public RequestTrieData(final byte[] nodeKey, final DatabaseType dbType, final int limit) {
+        super(Ver.V1, Ctrl.SYNC, Act.REQUEST_TRIE_DATA);
+
+        // ensure inputs are not null
+        Objects.requireNonNull(nodeKey);
+        Objects.requireNonNull(dbType);
+
+        // ensure limit is positive
+        if (limit < 0) {
+            throw new IllegalArgumentException(
+                    "The RequestTrieData object cannot be built with a negative limit.");
+        }
+
+        this.nodeKey = nodeKey;
+        this.dbType = dbType;
+        this.limit = limit;
+    }
+
+    /**
+     * Decodes a message into a trie node request.
+     *
+     * @param message a {@code byte} array representing a request for a trie node.
+     * @return the decoded trie node request if valid or {@code null} when the decoding encounters
+     *     invalid input
+     * @implNote Ensures that the components are not {@code null}.
+     */
+    public static RequestTrieData decode(final byte[] message) {
+        if (message == null || message.length == 0) {
+            return null;
+        } else {
+            RLPList list = (RLPList) RLP.decode2(message).get(0);
+            if (list.size() != 3) {
+                return null;
+            } else {
+                // decode the key
+                byte[] hash = list.get(0).getRLPData();
+                if (hash.length != 32) {
+                    return null;
+                }
+
+                // decode the db type
+                byte[] type = list.get(1).getRLPData();
+                DatabaseType dbType;
+                try {
+                    dbType = DatabaseType.valueOf(new String(type));
+                } catch (IllegalArgumentException e) {
+                    return null;
+                }
+
+                // decode the limit
+                int depth = new BigInteger(1, list.get(2).getRLPData()).intValue();
+
+                return new RequestTrieData(hash, dbType, depth);
+            }
+        }
+    }
+
+    @Override
+    public byte[] encode() {
+        return RLP.encodeList(
+                RLP.encodeElement(nodeKey),
+                RLP.encodeString(dbType.toString()),
+                RLP.encodeInt(limit));
+    }
+
+    /**
+     * Returns the blockchain database in which the requested key should be found.
+     *
+     * @return the blockchain database in which the requested key should be found
+     */
+    public DatabaseType getDbType() {
+        return dbType;
+    }
+
+    /**
+     * Returns the key of the requested trie node.
+     *
+     * @return the key of the requested trie node
+     */
+    public byte[] getNodeKey() {
+        return nodeKey;
+    }
+
+    /**
+     * Returns the maximum number of key-value pairs to be retrieved by the search inside the trie
+     * for referenced nodes, where:
+     *
+     * <ul>
+     *   <li>zero stands for searching for referenced nodes without a limit on the number of nodes
+     *       retrieved;
+     *   <li>one stands for not searching beyond the retrieved value for the given key;
+     *   <li>a positive value greater than one represents the number of additional key-value pairs
+     *       up to which to continue searching for referenced nodes.
+     * </ul>
+     *
+     * @return the maximum number of key-value pairs to be retrieved by the search inside the trie
+     *     for referenced nodes
+     */
+    public int getLimit() {
+        return limit;
+    }
+}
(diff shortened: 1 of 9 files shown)


modP2p/src/org/aion/p2p/Ver.java
```diff
@@ -8,16 +8,20 @@
 
     public static final short V0 = 0;
 
-    // for test
+    // new protocol version supporting fast sync
     public static final short V1 = 1;
 
+    // for test
+    public static final short V2 = 1;
```
AionJayT: change V2 to 2?

modAionImpl/src/org/aion/zero/impl/sync/msg/RequestTrieData.java
```diff
@@ -0,0 +1,134 @@
...
+        super(Ver.V1, Ctrl.SYNC, Act.REQUEST_TRIE_DATA);
+
+        // ensure inputs are not null
+        Objects.requireNonNull(nodeKey);
+        Objects.requireNonNull(dbType);
+
+        // ensure limit is positive
+        if (limit < 0) {
+            throw new IllegalArgumentException(
+                    "The RequestTrieData object cannot be built with a negative limit.");
```
aion-kelvin: "The RequestTrieData object must be built with a positive limit."

modAionImpl/src/org/aion/zero/impl/sync/msg/RequestTrieData.java
```diff
@@ -0,0 +1,134 @@
...
+     * @return the decoded trie node request if valid or {@code null} when the decoding encounters
+     *     invalid input
+     * @implNote Ensures that the components are not {@code null}.
+     */
+    public static RequestTrieData decode(final byte[] message) {
+        if (message == null || message.length == 0) {
+            return null;
+        } else {
+            RLPList list = (RLPList) RLP.decode2(message).get(0);
+            if (list.size() != 3) {
```
aion-kelvin: Is this an unexpected case?  If so, should it log an error/warning?

modAionImpl/src/org/aion/zero/impl/sync/msg/RequestTrieData.java
```diff
@@ -0,0 +1,134 @@
...
+                return null;
+            } else {
+                // decode the key
+                byte[] hash = list.get(0).getRLPData();
+                if (hash.length != 32) {
+                    return null;
```
aion-kelvin: same question re: logging

modAionImpl/src/org/aion/zero/impl/sync/msg/RequestTrieData.java
```diff
@@ -0,0 +1,134 @@
...
+                }
+
+                // decode the db type
+                byte[] type = list.get(1).getRLPData();
+                DatabaseType dbType;
+                try {
+                    dbType = DatabaseType.valueOf(new String(type));
+                } catch (IllegalArgumentException e) {
+                    return null;
```
aion-kelvin: same question re: logging

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
```diff
@@ -0,0 +1,259 @@
...
+        this.referencedNodes = referencedNodes;
+        this.dbType = dbType;
+    }
+
+    /**
+     * Decodes a message into a trie node response.
+     *
+     * @param message a {@code byte} array representing a response to a trie node request.
+     * @return the decoded trie node response
+     * @implNote The decoder must return {@code null} if any of the component values are {@code
```
aion-kelvin: Would it make sense to log something in case of component values being invalid?

Seems like otherwise it'd be really hard to debug if something in the kernel set up the input wrong earlier in the call stack

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
```diff
@@ -0,0 +1,259 @@
...
+                RLP.encodeList(encodeReferencedNodes(referencedNodes)),
+                RLP.encodeString(dbType.toString()));
+    }
+
+    /**
+     * Encodes each key-value pair from the {@link #referencedNodes} as a list of two elements.
+     *
+     * @return an array of all the key-value pair encodings
+     */
+    static byte[][] encodeReferencedNodes(Map<ByteArrayWrapper, byte[]> referencedNodes) {
```
aion-kelvin: Why package-visible?

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
```diff
@@ -0,0 +1,259 @@
...
+
+            // decode the key
+            hash = current.get(0).getRLPData();
+            if (hash.length != 32) {
+                return null;
+            }
+
+            // decode the value
+            value = current.get(1).getRLPData();
+            if (value.length == 0) {
```
aion-kelvin: could also do ! value.isEmpty()

modP2p/src/org/aion/p2p/Ver.java
```diff
@@ -8,16 +8,20 @@
 
     public static final short V0 = 0;
 
-    // for test
+    // new protocol version supporting fast sync
     public static final short V1 = 1;
 
+    // for test
+    public static final short VTEST = Byte.MAX_VALUE;
```
aion-kelvin: Could also use a negative number ... might be slightly clearer when looking at the actual number from a log file or something

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
```diff
@@ -0,0 +1,259 @@
...
+        if (message == null || message.length == 0) {
+            return null;
+        } else {
+            RLPList list = (RLPList) RLP.decode2(message).get(0);
+            if (list.size() != 4) {
+                return null;
+            } else {
+                // decode the key
+                byte[] hash = list.get(0).getRLPData();
+                if (hash.length != 32) {
```
aion-kelvin: might be worth putting 4, 32 in consts since they seem to show up all over the place

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * Decodes a message into a trie node response.
+     *
+     * @param message a {@code byte} array representing a response to a trie node request.
+     * @return the decoded trie node response
+     * @implNote The decoder must return {@code null} if any of the component values are {@code
+     *     null} or invalid.
+     */
+    public static ResponseTrieData decode(final byte[] message) {
+        if (message == null || message.length == 0) {
```
aion-kelvin: could also do message.isEmpty()

modP2p/src/org/aion/p2p/Ver.java
AlexandraRoatis: Done.

modAionImpl/src/org/aion/zero/impl/sync/msg/RequestTrieData.java
AlexandraRoatis: The handlers introduced in #817 and #818 will log errors in case decoding the messages was not possible. I don't think "the why" is meaningful. The essential here is to discard corrupt messages.

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
AlexandraRoatis: The unit tests cover various correct/incorrect scenarios for building the object. They would fail if the implementation changes to violate the expected behavior.

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
AlexandraRoatis: For testing. I will add the annotation.

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
AlexandraRoatis: The `value` is a `byte[]`. Don't have an `isEmpty()` method in this case.

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
AlexandraRoatis: The `message` is a `byte[]`. Don't have an `isEmpty()` method in this case.

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
AlexandraRoatis: I added `org.aion.p2p.V1Constants` for this purpose.

modP2p/src/org/aion/p2p/Ver.java
AlexandraRoatis: My intention here was only to not have to update the test version with every newly introduced protocol. Using negative values may have other ramifications that are not worth the trouble, given the fact that there exists only a minor test for this. If this starts being used for more substantial testing, the value can be updated to something more meaningful at that point.

modAionImpl/src/org/aion/zero/impl/sync/msg/ResponseTrieData.java
aion-kelvin: I meant if someone who is writing new code that calls ResponseTrieData.decode( .. ) -- they can still use it wrong and it wouldn't be covered by any existing test.  I don't debug this part of the code much, so I'll defer to your judgement ... it just seems like for someone who is implementing code that calls this function, or debugging a log file that touches this, having log messages for those unexpected cases would make debugging a lot easier if things go wrong.  I don't think there's any drawback to having logs at say ... debug or trace level, if polluting the output is a concern.

AlexandraRoatis: I don't think the place to log this information is inside the message, but I will expand the logged information done by the handlers in the next two PRs.

PR #817, opened 2019-02-07
Sync State 2: Handler for trie data request messages

AlexandraRoatis:
## Description

The PR contains all the functionality necessary for the trie data **request** handler to function correctly.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [x] Unit test.

## Testing

- existing tests + unit test for the handler

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/sync/handler/RequestTrieDataHandler.java
+++ b/modAionImpl/src/org/aion/zero/impl/sync/handler/RequestTrieDataHandler.java
@@ -1,5 +1,7 @@
 package org.aion.zero.impl.sync.handler;
 
+import static org.aion.p2p.V1Constants.TRIE_DATA_REQUEST_MAXIMUM_BATCH_SIZE;
+
 import java.util.Map;
 import org.aion.base.util.ByteArrayWrapper;
 import org.aion.p2p.Ctrl;
@@ -25,9 +27,6 @@ public final class RequestTrieDataHandler extends Handler {
     private final IAionBlockchain chain;
 
     private final IP2pMgr p2p;
-
-    // limits the number of key-value pairs returned to one request
-    public static final int MAXIMUM_BATCH_SIZE = 100;
 
     /**
      * Constructor.
@@ -74,11 +73,14 @@ public final class RequestTrieDataHandler extends Handler {
                 } else {
                     // check for internal limit on the request
                     if (limit == 0) {
-                        limit = MAXIMUM_BATCH_SIZE;
+                        limit = TRIE_DATA_REQUEST_MAXIMUM_BATCH_SIZE;
                     } else {
                         // the first value counts towards the limit
                         limit--;
-                        limit = limit < MAXIMUM_BATCH_SIZE ? limit : MAXIMUM_BATCH_SIZE;
+                        limit =
+                                limit < TRIE_DATA_REQUEST_MAXIMUM_BATCH_SIZE
+                                        ? limit
+                                        : TRIE_DATA_REQUEST_MAXIMUM_BATCH_SIZE;
                     }
 
                     // determine if the node can be expanded
--- a/modAionImpl/src/org/aion/zero/impl/sync/handler/RequestTrieDataHandler.java
+++ b/modAionImpl/src/org/aion/zero/impl/sync/handler/RequestTrieDataHandler.java
@@ -2,6 +2,7 @@ package org.aion.zero.impl.sync.handler;
 
 import static org.aion.p2p.V1Constants.TRIE_DATA_REQUEST_MAXIMUM_BATCH_SIZE;
 
+import java.util.Arrays;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


modMcf/src/org/aion/mcf/trie/TrieImpl.java
```diff
@@ -881,6 +882,64 @@ private void traceTrie(byte[] stateRoot, ScanAction action) {
...
+    public Map<ByteArrayWrapper, byte[]> getReferencedTrieNodes(byte[] value, int limit) {
+        CollectMappings collect = new CollectMappings();
+        ArrayList<byte[]> hashes = new ArrayList<>();
+        Value node = Value.fromRlpEncoded(value);
+
+        synchronized (cache) {
+            appendHashes(node, hashes);
+
+            while (!hashes.isEmpty() && collect.getSize() < limit) {
+                byte[] myHash = hashes.remove(0);
```
AionJayT: just think about you could avoid the remove call. Considering use the following code snipet and add element during the for loop might be faster.
https://stackoverflow.com/questions/993025/adding-elements-to-a-collection-during-iteration
```
List<ZeObj> myList = new ArrayList<ZeObj>();
    // populate the list with whatever
            ........
    int noItems = myList.size();
    for (int i = 0; i < noItems; i++) {
        ZeObj currItem = myList.get(i);
        // when you want to add, simply add the new item at last and
        // increment the stop condition
        if (currItem.asksForMore()) {
            myList.add(new ZeObj());
            noItems++;
        }
    }
```

modAionImpl/src/org/aion/zero/impl/sync/handler/RequestTrieDataHandler.java
```diff
@@ -0,0 +1,110 @@
...
+                    // determine if the node can be expanded
+                    Map<ByteArrayWrapper, byte[]> referencedNodes =
+                            chain.getReferencedTrieNodes(value, limit, dbType);
+
+                    // generate response with referenced nodes
+                    response = new ResponseTrieData(key, value, referencedNodes, dbType);
+                }
+
+                // reply to request
+                this.p2p.send(peerId, displayId, response);
```
AionJayT: should we do try-catch before p2p.send? Does any place catch the throws from getTrieNode and getReferencedTrieNodes?

modMcf/src/org/aion/mcf/trie/TrieImpl.java
```diff
@@ -881,6 +882,64 @@ private void traceTrie(byte[] stateRoot, ScanAction action) {
...
+        } else if (node.isList()) {
+            List<Object> siblings = node.asList();
+            if (siblings.size() == PAIR_SIZE) {
+                Value val = new Value(siblings.get(1));
+                if (val.isHashCode() && !hasTerminator((byte[]) siblings.get(0))) {
+                    hashes.add(val.asBytes());
+                }
+            } else {
+                for (int j = 0; j < LIST_SIZE; ++j) {
+                    Value val = new Value(siblings.get(j));
```
aion-kelvin: Is it possible for siblings.size() to be neither PAIR_SIZE nor LIST_SIZE?

I'm just thinking, if siblings.size() is 3, then line 934 will get array out of bounds exception.

AlexandraRoatis: It's not possible to have list nodes with different sizes other than the two above.

modAionImpl/src/org/aion/zero/impl/sync/handler/RequestTrieDataHandler.java
```diff
@@ -0,0 +1,110 @@
...
+
+            if (value != null) {
+                ResponseTrieData response;
+
+                if (limit == 1) {
+                    // generate response without referenced nodes
+                    response = new ResponseTrieData(key, value, dbType);
+                } else {
+                    // check for internal limit on the request
+                    if (limit == 0) {
```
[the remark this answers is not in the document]
AlexandraRoatis: Since this is encoded using RLP I could not use negative values for signaling `unlimited`, but allowing the receiver to limit the size of the output sent back is a useful case. Further, it doesn't make sense to submit a request over the network that returns no data, so the value 0 is not meaningful in the context of the request and can be used for a different purpose.

On the other hand, `getReferencedTrieNodes` can be a very expensive operation and I preferred to use the limit literally and not allow unlimited searches due to the potential of accidental or purposeful misuse. There's the option of throwing an IllegalArgumentException for values <= 0, but I think returning an empty set is preferable.

modAionImpl/src/org/aion/zero/impl/sync/handler/RequestTrieDataHandler.java
AlexandraRoatis: I think there were enough parameter validations to avoid the errors, but I've added the `try-catch` just in case something was missed because the handler should continue working even if some retrieval methods fail.

PR #820, opened 2019-02-11
Sync State 4: Task for importing trie data

AlexandraRoatis:
## Description

The PR contains the functionality necessary for the trie data received in the responses to be imported into the repository database.

It only contains a skeleton of the `FastSyncManager.java` which will be completed in subsequent PRs.

## Type of change

- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [x] Unit test.

## Testing

- existing tests + unit test for the handler

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
@@ -29,6 +29,7 @@ import org.aion.mcf.trie.SecureTrie;
 import org.aion.mcf.trie.Trie;
 import org.aion.mcf.trie.TrieImpl;
 import org.aion.mcf.trie.TrieNodeResult;
+import org.aion.p2p.V1Constants;
 import org.aion.vm.api.interfaces.Address;
 import org.aion.zero.db.AionRepositoryCache;
 import org.aion.zero.impl.config.CfgAion;
@@ -890,7 +891,7 @@ public class AionRepositoryImpl
         }
 
         // empty keys are not allowed
-        if (key == null || key.length != 32) {
+        if (key == null || key.length != V1Constants.HASH_SIZE) {
             return TrieNodeResult.INVALID_KEY;
         }
 
--- a/modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
@@ -28,6 +28,7 @@ import org.aion.mcf.db.TransactionStore;
 import org.aion.mcf.trie.SecureTrie;
 import org.aion.mcf.trie.Trie;
 import org.aion.mcf.trie.TrieImpl;
+import org.aion.mcf.trie.TrieNodeResult;
 import org.aion.vm.api.interfaces.Address;
 import org.aion.zero.db.AionRepositoryCache;
 import org.aion.zero.impl.config.CfgAion;
@@ -864,6 +865,42 @@ public class AionRepositoryImpl
             Trie trie = new TrieImpl(db);
             return trie.getReferencedTrieNodes(value, limit);
         }
+    }
+
+    /**
+     * Imports a trie node to the indicated blockchain database.
+     *
+     * @param key the hash key of the trie node to be imported
+     * @param value the value of the trie node to be imported
+     * @param dbType the database where the key-value pair should be stored
+     * @throws IllegalArgumentException if the given key is null or the database type is not
+     *     supported
+     * @return a {@link TrieNodeResult} indicating the success or failure of the import operation
+     */
+    public TrieNodeResult importTrieNode(byte[] key, byte[] value, DatabaseType dbType) {
+        IByteArrayKeyValueDatabase db = selectDatabase(dbType);
+
+        Optional<byte[]> stored = db.get(key);
+        if (stored.isPresent()) {
+            if (Arrays.equals(stored.get(), value)) {
+                return TrieNodeResult.KNOWN;
+            } else {
+                return TrieNodeResult.INCONSISTENT;
+            }
+        }
+
+        // empty keys are not allowed
+        if (key == null || key.length != 32) {
+            return TrieNodeResult.INVALID_KEY;
+        }
+
+        // not allowing deletions to be imported
+        if (value == null || value.length == 0) {
+            return TrieNodeResult.INVALID_VALUE;
+        }
+
+        db.put(key, value);
+        return TrieNodeResult.IMPORTED;
     }
 
     private IByteArrayKeyValueDatabase selectDatabase(DatabaseType dbType) {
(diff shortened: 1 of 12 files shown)


modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -866,6 +868,42 @@ public void compactState() {
...
+
+        // empty keys are not allowed
+        if (key == null || key.length != V1Constants.HASH_SIZE) {
+            return TrieNodeResult.INVALID_KEY;
+        }
+
+        // not allowing deletions to be imported
+        if (value == null || value.length == 0) {
+            return TrieNodeResult.INVALID_VALUE;
+        }
```
AionJayT: why not check the key value in the beginning?

AlexandraRoatis: good point, I'll update it

modAionImpl/src/org/aion/zero/impl/sync/TaskImportTrieData.java
```diff
@@ -0,0 +1,141 @@
...
+     * already imported data.
+     *
+     * @param wrapper the initial set of trie nodes to be imported
+     * @return the remaining nodes after the exact matches have been filtered out
+     */
+    private Map<ByteArrayWrapper, byte[]> filterImported(TrieNodeWrapper wrapper) {
+        Map<ByteArrayWrapper, byte[]> nodes =
+                wrapper.getReferencedNodes()
+                        .entrySet()
+                        .parallelStream()
```
AionJayT: how many nodes do you expect during the fast-sync? 
http://gee.cs.oswego.edu/dl/html/StreamParallelGuidance.html

modAionImpl/src/org/aion/zero/impl/sync/TaskImportTrieData.java
```diff
@@ -0,0 +1,141 @@
...
+                        .filter(e -> !fastSyncMgr.containsExact(e.getKey(), e.getValue()))
+                        .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+
+        if (!fastSyncMgr.containsExact(wrapper.getNodeKey(), wrapper.getNodeValue())) {
```
AionJayT: any reason for doing the fastSyncMgr.containsExact() twice? I think the stream() already checked the logic.

modAionImpl/test/org/aion/zero/impl/db/AionRepositoryImplTest.java
```diff
@@ -371,4 +375,50 @@ public void testGetSnapshotToRoot() {
...
+
+    @Test
+    public void testImportTrieNode() {
+        AionRepositoryImpl repository = AionRepositoryImpl.createForTesting(repoConfig);
+        IByteArrayKeyValueDatabase db = repository.getStateDatabase();
+        byte[] nodeKey = RequestTrieDataTest.nodeKey;
+        byte[] altNodeKey = RequestTrieDataTest.altNodeKey;
+        byte[] smallNodeKey = RequestTrieDataTest.smallNodeKey;
+        byte[] leafValue = ResponseTrieDataTest.leafValue;
+        byte[] branchValue = ResponseTrieDataTest.branchValue;
```
aion-kelvin: Might be safer to just duplicate those values here instead of referring to them (I think if someone is updating ResponseTriedatatest.branchValue they might not expect that some other test also uses it)

modAionImpl/src/org/aion/zero/impl/sync/FastSyncManager.java
```diff
@@ -0,0 +1,138 @@
...
+    public boolean containsExact(ByteArrayWrapper key, byte[] value) {
+        return enabled
+                && importedTrieNodes.containsKey(key)
+                && Arrays.equals(importedTrieNodes.get(key), value);
+    }
+
+    /** Changes the pivot in case of import failure. */
+    public void handleFailedImport(
+            ByteArrayWrapper key, byte[] value, DatabaseType dbType, int peerId, String peer) {
+        if (enabled) {
```
aion-kelvin: Is there any valid reason to call this while enabled is false?

modAionImpl/src/org/aion/zero/impl/sync/TaskImportTrieData.java
```diff
@@ -0,0 +1,141 @@
...
+        // since it is usually the longest process (on large blockchains)
+        Thread.currentThread().setPriority(Thread.MAX_PRIORITY);
+
+        while (!fastSyncMgr.isComplete()) {
+            TrieNodeWrapper tnw;
+            try {
+                tnw = trieNodes.take();
+            } catch (InterruptedException ex) {
+                if (!fastSyncMgr.isComplete()) {
+                    log.error("<import-trie-nodes: interrupted without shutdown request>", ex);
```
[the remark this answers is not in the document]
AionJayT: good point!!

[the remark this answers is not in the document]
AlexandraRoatis: Similar to `TaskImportBlocks.java` the sync manager will know of an instance of this class and trigger the shutdown when appropriate.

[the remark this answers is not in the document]
aion-kelvin: I see... is that shutdown just setting fastSyncMgr.isComplete to true?

modAionImpl/src/org/aion/zero/impl/sync/FastSyncManager.java
AlexandraRoatis: The functionality is not fully defined there yet so I'm not entirely sure. I came across potential un-recoverable errors that will require disabling fast sync. That `if` is there more as a reminer that some methods may still be called while the message gets through to all threads. This class can be ignored for the moment, since the methods in it are now implemnted yet.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportTrieData.java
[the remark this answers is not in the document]
AlexandraRoatis: Yep. Since the functionality is a one-time catch-up with the chain use case, it will be permanently disabled after completion. In fact, it also will not be enabled if the initial database is not empty.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportTrieData.java
AlexandraRoatis: The current limit is set to 100 entries, so I will remove the parallel stream.

modAionImpl/src/org/aion/zero/impl/sync/TaskImportTrieData.java
AlexandraRoatis: The second call is applied to different values. The wrapper contains a key and its corresponding value that were requested, plus a set of referenced key-value pairs (other key-value entries that the value above points to).

modAionImpl/test/org/aion/zero/impl/db/AionRepositoryImplTest.java
AlexandraRoatis: Done.

PR #826, opened 2019-02-14
Enable AVM inside BulkExecuter based on given config

AlexandraRoatis:
## Description

Ensures the AVM is used only when enabled.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

- existing tests + sync on `mastery`

## Verification

### DIFF
--- a/modVM/src/org/aion/vm/BulkExecutor.java
+++ b/modVM/src/org/aion/vm/BulkExecutor.java
@@ -55,6 +55,12 @@ import org.slf4j.Logger;
 public class BulkExecutor {
     private static final Object LOCK = new Object();
 
+    private static boolean avmEnabled = false;
+
+    public static void enabledAvmCheck(boolean isEnabled) {
+        avmEnabled = isEnabled;
+    }
+
     private IRepository repository;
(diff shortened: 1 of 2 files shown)


modVM/src/org/aion/vm/BulkExecutor.java
```diff
@@ -55,6 +55,12 @@
 public class BulkExecutor {
     private static final Object LOCK = new Object();
 
+    private static boolean avmEnabled = false;
+
+    public static void enabledAvmCheck(boolean isEnabled) {
```
aionick: Not a big fan of this means of passing the value. It puts the burden of knowing how to use an internal detail on the caller, when the bulk executor can grab this answer itself internally without any assistance from the config. A detail like this will eventually leak into unit tests, for example.

AlexandraRoatis: I understand your point of view, but the caller should be in charge of directing behavior here. If every class queries the config when it chooses there is more potential for errors (especially when you consider that the config is malleable and something like the GUI may choose to edit it) than if one entity directs overall behavior. The tests should be written to consider both AVM enabled and disabled.

PR #830, opened 2019-02-19
added more logging info for test harness usage

aionWilliam:
## Notice

## Description
added more logging info for test harness usage 

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

-

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
@@ -100,6 +100,7 @@ import org.slf4j.LoggerFactory;
 public class AionBlockchainImpl implements IAionBlockchain {
 
     private static final Logger LOG = LoggerFactory.getLogger(LogEnum.CONS.name());
+    private static final Logger TX_LOG = LoggerFactory.getLogger(LogEnum.TX.name());
     private static final int THOUSAND_MS = 1000;
     private static final int DIFFICULTY_BYTES = 16;
 
@@ -664,6 +665,20 @@ public class AionBlockchainImpl implements IAionBlockchain {
             }
         }
 
+        if (ret == IMPORTED_BEST) {
+            if (TX_LOG.isDebugEnabled()) {
+                if (summary != null){
+                    for (AionTxReceipt receipt : summary.getReceipts()) {
+                        if (receipt != null) {
+                            byte[] transactionHash = receipt.getTransaction().getTransactionHash();
+                            TX_LOG.debug("Transaction: " + Hex.toHexString(transactionHash) +
+                                    " was sealed into block #" + block.getNumber());
+                        }
+                    }
+                }
+            }
+        }
+
         return Pair.of(ret, summary);
     }
 
@@ -1032,9 +1047,9 @@ public class AionBlockchainImpl implements IAionBlockchain {
                                                 || !TransactionTypeValidator.isValid(
                                                         tx.getTargetVM()))) {
                     LOG.error("Some transactions in the block are invalid");
-                    if (LOG.isDebugEnabled()) {
+                    if (TX_LOG.isDebugEnabled()) {
                         for (AionTransaction tx : txs) {
-                            LOG.debug(
+                            TX_LOG.debug(
                                     "Tx valid ["
                                             + TXValidator.isValid(tx)
                                             + "]. Type valid ["
--- a/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
@@ -667,12 +667,15 @@ public class AionBlockchainImpl implements IAionBlockchain {
 
         if (ret == IMPORTED_BEST) {
             if (TX_LOG.isDebugEnabled()) {
-                if (summary != null){
+                if (summary != null) {
                     for (AionTxReceipt receipt : summary.getReceipts()) {
                         if (receipt != null) {
                             byte[] transactionHash = receipt.getTransaction().getTransactionHash();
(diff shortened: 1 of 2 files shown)


modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -664,6 +665,15 @@ public synchronized void compactState() {
             }
         }
 
+        if (ret == IMPORTED_BEST) {
+            if (TX_LOG.isDebugEnabled()) {
+                for (AionTxReceipt receipt : summary.getReceipts()) {
+                    byte[] transactionHash = receipt.getTransaction().getTransactionHash();
+                    TX_LOG.debug("Transaction: " + Hex.toHexString(transactionHash) + " was sealed into block #" + block.getNumber());
```
aion-kelvin: do we need to do null checks on summary, receipt, and receipt.getTransaction() ?

don't want to accidentally throw an NPE while trying to construct the log message.

aionWilliam: > Could you add the info from line 1045 to that log as well?
> Also please update the classes to use the [correct formatting](https://github.com/aionnetwork/aion/wiki/Aion-Code-Conventions).

Thanks for the comment @AlexandraRoatis, I am assuming you mean line 1045 in AionBlockchainImpl? Im wondering why do we need this, cuz I dont think the conditions match?
https://github.com/aionnetwork/aion/blob/0ed32f107c71e255561dfadffe1f794426bcbf07/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java#L1045

aionWilliam:
> Could you add the info from line 1045 to that log as well?
> Also please update the classes to use the [correct formatting](https://github.com/aionnetwork/aion/wiki/Aion-Code-Conventions).

Thanks for the comment @AlexandraRoatis, I am assuming you mean line 1045 in AionBlockchainImpl? Im wondering why do we need this, cuz I dont think the conditions match?
https://github.com/aionnetwork/aion/blob/0ed32f107c71e255561dfadffe1f794426bcbf07/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java#L1045

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
aionWilliam: Thanks for the comment @aion-kelvin, for summary the NPE is already checked earlier in the method. Correct me if im wrong, as for tx and receipt I dont believe they would be throwing NPE at this stage, and if they do throw NPE, we should probably catch it.

aion-kelvin: You meant 'returns null' when you wrote 'throwing NPE' right?

Line 634 sets summary to null and then, as far as I can see, proceeds to your newly-added code without setting it to anything else (note that after Line 634, the only thing that uses summary is after a check on ret; whereas your code will try to use summary regardless of what ret was).

As for receipt and receipt.getTransaction -- I am not totally sure what they do.  I think you're right that they're not supposed to be returning null at this stage.  But bugs happen (especially since in our code we have a habit of returning null when things don't work or things can't be found) -- it would be nice to be defensive against that.

AlexandraRoatis:
@aionWilliam since we now have a dedicated log for transactions, I think we should add the transaction-related information to it. We'll probably move more logs to it later on. The message at line 1045 was very useful recently and I thought it was worth moving it to the new log.

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
aionWilliam: Okay, I've added the null checkers, I agree its always nice to do defensive programming.

aionWilliam:
@AlexandraRoatis I've added those validation logs for transactions. I agree they would be pretty useful.

modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
```diff
@@ -714,8 +714,8 @@ private TxResponse addPendingTransactionImpl(final AionTransaction tx, BigIntege
         if (txSum.isRejected()) {
             if (LOGGER_TX.isTraceEnabled()) {
                 LOGGER_TX.trace(
-                        "addPendingTransactionImpl tx is rejected due to: {}",
-                        txSum.getReceipt().getError());
+                        "addPendingTransactionImpl tx " + Hex.toHexString(tx.getTransactionHash())
+                                + " is rejected due to: {}", txSum.getReceipt().getError());
```
AlexandraRoatis: incorrect formatting

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -664,6 +665,20 @@ public synchronized void compactState() {
...
         }
 
+        if (ret == IMPORTED_BEST) {
+            if (TX_LOG.isDebugEnabled()) {
+                if (summary != null){
+                    for (AionTxReceipt receipt : summary.getReceipts()) {
+                        if (receipt != null) {
+                            byte[] transactionHash = receipt.getTransaction().getTransactionHash();
+                            TX_LOG.debug("Transaction: " + Hex.toHexString(transactionHash) +
+                                    " was sealed into block #" + block.getNumber());
```
AlexandraRoatis: incorrect formatting

aionWilliam: They are fixed now.

PR #845, opened 2019-03-12
Fix to eth_syncing

aionick:
## Description

- The eth_syncing method was reporting the opposite of what it should have been: if the local node was less than 5 blocks of the network best it was considered "done syncing" (hence if it was at block 3 of 500, it said 'done'), and if it was within 5 blocks it was considered "still syncing". This fixes that logic.
- Added some unit tests to cover the testable part of the logic: `ApiAion::getSyncInfo`.
- Removed the `isSyncComplete` method that `eth_syncing` used to rely on. There is no longer any need for this method. It also grabbed the local & network best blocks on its own, disregarding the fact that it's caller could be using different values.
- Removed 2 tests on the `isSyncComplete` method. I could not determine what these tests were verifying and whether it was relevant. We now have tests on `ApiAion::getSyncInfo` which is what both API's now make use of, so this should cover the removed tests anyway.
- The API calls into eth_syncing are more well-defined now: they now handle the error cases of not knowing the local or network block numbers instead of giving these default values and pretending like  nothing happened. This should also make it easier now for a client to determine whether or not they've connected to the network.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- Tested out extensively in the node test harness framework.
- Added a few unit tests as well.

## Verification


aion-kelvin:
> Added a new interpretation to "still syncing": if the network best block is zero then you are still syncing. Why? Because you are still trying to connect to your peers.

I believe `eth_syncing` is supposed to return either "false" if _it is not syncing_, or the information about the sync if it is syncing.

In the case of best blocking being zero, I think it should report that it is not syncing, instead of saying that it is "still syncing."

Is that possible?

AionJayT:
We probably can consider an additional case. Say if I am trying to sync to the network but I have a problem after synced 500 blocks then the kernel stuck. Should we consider it as a not syncing case?

aionick:
The confusion here was caused by the `isSyncComplete()` method, whose implementation is completely wrong but it becomes correct because the calling class uses it as if it is named `isStillSyncing()`. I will get around to fixing up this PR accordingly soon.

aion-kelvin:
@aionick So, to follow up on our earlier chat about the correct behaviour of eth_syncing, I chatted a bit with @AlexandraRoatis and I propose doing this:

- we make it return "false" if bestBlock == currentBlock.  This also covers the case of having zero peers because in that case, bestBlock is just your currentBlock.  (If needed, user can verify that there are peers using `net_peerCount`).
- in all other cases, return the struct.  

I think this makes the most sense because you care about the sync progress when you're NOT synched, not the other way around.  

The way to interpret the "false" would be that syncing is not happening because we're already at the best block. 

Thoughts?

aionick:
>  This also covers the case of having zero peers because in that case, bestBlock is just your currentBlock.
Do you mean bestBlock here as in the network best? If so, this isn't necessarily true. I can be at block N > 1 on a standalone node and since I have no peers the network best block is 0. You can only detect this case by checking network best is zero, and you can't distinguish between standalone or trying to connect.

But yes I agree on the interpretation of false

aion-kelvin:
> I can be at block N > 1 on a standalone node and since I have no peers the network best block is 0.

Oh, interesting.

What if we made it:

- false if currentBlock >= networkBestBlock
- struct otherwise

aionick:
For the testing harness, I would prefer if we could interpret "best block == 0" as "still syncing", it greatly simplifies the logic. This is only "wrong" in a standalone situation, but you should not even be checking if you are synced in a standalone situation anyway, because that's trivially true (or arguably undefined)

AlexandraRoatis:
I don't really agree with that interpretation. The actual value of the best block should not muddle the definition of is syncing or not. Checking the number of peers can be used for the above use case. If there are peers in the boot list but no active connections, it can be interpreted as still syncing for testing purposes.
I think of unfortunate side effects coming from this change in interpretation. Say you define the condition for mining a block to be synced to avoid producing side chains, now you can't start mining if you're at Genesis because of changes made to facilitate testing.
Additionally, I think we should be using the harness to test real situations. Checking for peers is part of a real scenario. A developer would definitely ensure this while testing the kernel, and I feel part of the purpose for the harness is to automatize what would otherwise need a dev to do.

aion-kelvin:
The `"best block == 0" ==> "still syncing"` interpretation is slightly odd, but I could accept it as a compromise.

@AlexandraRoatis brings up some good points though and I think it is a more disciplined approach.

If the extra work for the test harness isn't huge, I think it would be worth it to do as she suggests.  From what you were telling me, Nick, I think the extra work is that at kernel start-up time, you'd have to poll for peers via net_peerCount before doing the main work of the test.  While a little cumbersome, I feel like the pattern of "keep polling kernel until some property has changed" is something that will be used a lot in the testing, so building that out might be actually pretty useful.

aionick:
Alright makes sense, I suppose we should do things properly... ugh :P

aionick:
This PR is still on haitus for a bit. Checking over the specification again, `eth_syncing` is incorrect and these changes are necessary. I will make the suggested changes @AlexandraRoatis and @aion-kelvin mentioned once I get back to this. The test harness will also need to change how it does things slightly as a result, so I'll be doing that as well.

aionick:
This is ready to be reviewed again

modApiServer/src/org/aion/api/server/ApiAion.java
```diff
@@ -50,6 +50,7 @@
 import org.aion.zero.types.AionTxReceipt;
 
 public abstract class ApiAion extends Api {
+    public static final int SYNC_TOLERANCE = 5;
```
aion-kelvin: Is there any reason not to just make this 1?

aionick: I guess realistically you can be within ~50 blocks and you will be safe to send a transaction to the node because by the time you do you will be at the top, so I don't think there's much of a difference for any value below this point, and especially for any values below 10.

Since the old value was 5 I don't think there's much point in changing it really? But if you guys want to see this change to 1 that's probably fine.. though we should run the test harness against that value just to ensure there's no weirdness from how close this value is to zero. Just for this extra work I feel we should leave it be since there's no practical difference.

aion-kelvin: It's not a huge deal (especially considering this function did the opposite of what it was supposed to for so long)... it's just -- when it's 1, it's easier for the user to figure out network best block. Because if you eth_syncing is returning false, you know that networkBest=localBest, so you can eth_getBlockNumber and know what localBest and networkBest are.  Whereas if it's 5, then when it returns false, the you're only able to deduce that networkBest is in [localBest-5,localBest].

aionick: Fair enough. I will hopefully get around to testing this out against the harness soon just to see that it behaves okay and that 1 isn't an issue. I don't suspect it will be.

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -96,6 +97,8 @@
 @SuppressWarnings({"Duplicates", "WeakerAccess"})
 public class ApiWeb3Aion extends ApiAion {
 
+    private static final int SYNC_TOLERANCE = 5;
```
AlexandraRoatis: should this also be set to 1?

PR #848, opened 2019-03-14
Avm now handles "pure" balance transfers

aionick:
## Description

- Formerly, transactions that simply transferred balance to an address that is not a contract were executed by the Fvm. Now they are executed by the Avm.
- This now mean that a CALL transaction is avm-bound if: the call is to an avm contract _or to a non-contract address_.
- Some minor refactorings to the `BulkExecutor`; namely, removing the unnecessary `isTransactionForFvm()` method which must simply return `!isTransactionForAvm()`.
- There is a discrepancy between vm's handling the pay miner fees & refund sender. The Avm does this whereas the Fvm does not (will open an issue for this soon). This explains the other logic change in the `BulkExecutor`, which only does these actions if the transaction is not avm-bound.

**The actual changes to logic are in commit b177d64 (the other commit, b61bc8c contains tests).**

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- A number of consensus tests were added before making the change, in the first commit here, to ensure consensus was not broken. The test harness also contains some tests on balance transfers (soon to be pushed)

## Verification


modVM/src/org/aion/vm/BulkExecutor.java
```diff
@@ -124,7 +123,7 @@ public BulkExecutor(
      * @param logger The logger.
      * @param work The post-execution work to apply after each transaction is run.
      */
-    public BulkExecutor (
+    public BulkExecutor(
```
AlexandraRoatis: formatting

modVM/src/org/aion/vm/BulkExecutor.java
```diff
@@ -343,25 +345,25 @@ private void updateRepository(
...
             // Refund energy if transaction was successfully or reverted.
-            if (result.getResultCode().isSuccess() || result.getResultCode().isRevert()) {
-                if (!transactionIsForAionVirtualMachine(tx)) {
+            if (!transactionIsForAionVirtualMachine(tx)) {
+                if (result.getResultCode().isSuccess() || result.getResultCode().isRevert()) {
                     track.addBalance(tx.getSenderAddress(), summary.getRefund());
                 }
-            }
 
-            tx.setNrgConsume(computeEnergyUsed(tx.getEnergyLimit(), result));
```
AlexandraRoatis: where is this done now?

aionick: It just moved up a few lines, before the if statement, since it is not conditional

aionick:
I won't have time to do the assertThat change, I'll try to keep it in mind in the future

modVM/src/org/aion/vm/BulkExecutor.java
```diff
@@ -157,45 +156,47 @@ public BulkExecutor (
...
+                            new KernelInterfaceForFastVM(
                                     this.repositoryChild.startTracking(),
                                     this.allowNonceIncrement,
                                     this.isLocalCall);
                     virtualMachineForNextBatch =
-                            VirtualMachineProvider.getVirtualMachineInstance(VM.AVM, vmKernel);
+                            VirtualMachineProvider.getVirtualMachineInstance(VM.FVM, vmKernel);
                     nextBatchToExecute =
-                            fetchNextBatchOfTransactionsForAionVirtualMachine(currentIndex);
+                            fetchNextBatchOfTransactionsForFastVirtualMachine(currentIndex);
```
AionJayT: reason for reversing avm fvm order in the if-else condition

aionick: Just seems more natural to do `if (condition)` instead of `if (!condition)`

AionJayT:
please fix the conflict. Thanks

modVM/src/org/aion/vm/BulkExecutor.java
aionick: I don't think that's a formatting mistake. That's the result of applying the formatter, lol.

AlexandraRoatis: ups, sorry about that!

PR #866, opened 2019-04-04
Contract storage enhancement for use of object graph

AlexandraRoatis:
## Description

The major change of this PR is allowing a **different interpretation of the `storageRoot`** for FVM contracts (same as before) and AVM contracts. 

In the case of **AVM contracts** the **storage root is a pointer** to a database value that retrieves the long term storage root and the hash of the object graph. The reason for overwriting the meaning of the `storageRoot` is to maintain the same number of elements of an account involved in consensus.

To achieve this change the following classes were updated in this PR:
* `KernelInterfaceForAVM.jave` implements the put/get object graph methods by delegating the call to the repository.
* `KernelInterfaceForFastVM.java` does not support the object graph and therefore throws an exception when either method is called.
* `BulkExecutor.java` can determine the VM type by either calling the contract through a copy of the repository or by making a direct DB call as before.
* `AionBlockchainImpl.java` also save the VM type inside the contract details object.
* `CfgDb.java`, `AionRepositoryImpl.java`, `AbstractRepository.java`, `AionRepositoryCache.java` and `DetailsDataStore.java` have a new database for storing the object graph and new method for setting and retrieving the VM type and object graph.
* `AbstractContractDetails.java`, `AionContractDetailsImpl.java`, `ContractDetailsCacheImpl.java` include tracking of the VM type and to different degrees the object graph. Because the AVM implementation is not yet updated to use the object graph a boolean flag `enableObjectGraph` disables some of the functionality. The encoding of the contract details now contains also the VM type. When the contract details are created as a snapshot to a specific storage hash, they must search the database for the actual storage root and object graph hash.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- existing tests

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
@@ -1342,6 +1342,18 @@ public class AionBlockchainImpl implements IAionBlockchain {
             for (AionTxExecSummary summary : executionSummaries) {
                 receipts.add(summary.getReceipt());
                 summaries.add(summary);
+
+                // save contract creation data to contract details
+                AionTxReceipt receipt = summary.getReceipt();
+                AionTransaction tx = receipt.getTransaction();
+                if (tx.isContractCreationTransaction() && receipt.isSuccessful()) {
+                    track.saveVmType(
+                            tx.getContractAddress(),
+                            TransactionTypeRule.isValidAVMContractDeployment(tx.getTargetVM())
+                                    ? TransactionTypes.AVM_CREATE_CODE
+                                    // FVM contracts do not always have the correct type
+                                    : TransactionTypes.FVM_CREATE_CODE);
+                }
             }
         }
         Map<Address, BigInteger> rewards = addReward(block);
--- a/modAionImpl/src/org/aion/zero/impl/db/AbstractContractDetails.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AbstractContractDetails.java
@@ -23,8 +23,11 @@ public abstract class AbstractContractDetails implements ContractDetails {
     protected int detailsInMemoryStorageLimit;
 
     private Map<ByteArrayWrapper, byte[]> codes = new HashMap<>();
-    // set to FVM_CREATE_CODE to update encoding for FVM contracts
-    protected byte vmType = TransactionTypes.FVM_CREATE_CODE;
+    // classes extending this rely on this value starting off as null
+    protected byte[] objectGraph = null;
+
+    // using the default transaction type to specify undefined VM
+    protected byte vmType = TransactionTypes.DEFAULT;
 
     protected AbstractContractDetails() {
         this(0, 64 * 1024);
@@ -73,14 +76,6 @@ public abstract class AbstractContractDetails implements ContractDetails {
 
     public void appendCodes(Map<ByteArrayWrapper, byte[]> codes) {
         this.codes.putAll(codes);
-    }
-
-    public void setVmType(byte vmType) {
-        this.vmType = vmType;
-    }
-
-    public byte getVmType() {
-        return vmType;
     }
 
     @Override
--- a/modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
@@ -130,6 +130,14 @@ public class AionContractDetailsImpl extends AbstractContractDetails {
     }
 
     @Override
+    public void setCode(byte[] code) {
+        super.setCode(code);
+        if (isDirty()) {
+            rlpEncoded = null;
+        }
+    }
+
+    @Override
     public byte[] getObjectGraph() {
         if (objectGraph == null) {
             // the object graph was not stored yet
--- a/modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
@@ -1,7 +1,9 @@
 package org.aion.zero.impl.db;
 
+import static org.aion.crypto.HashUtil.EMPTY_DATA_HASH;
 import static org.aion.crypto.HashUtil.EMPTY_TRIE_HASH;
 import static org.aion.crypto.HashUtil.h256;
+import static org.aion.mcf.tx.TransactionTypes.FVM_CREATE_CODE;
 import static org.aion.types.ByteArrayWrapper.wrap;
 import static org.aion.util.bytes.ByteUtil.EMPTY_BYTE_ARRAY;
 
@@ -11,11 +13,13 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.Optional;
 import org.aion.interfaces.db.ByteArrayKeyValueStore;
 import org.aion.interfaces.db.ContractDetails;
 import org.aion.mcf.ds.XorDataSource;
 import org.aion.mcf.trie.SecureTrie;
 import org.aion.mcf.tx.TransactionTypes;
+import org.aion.precompiled.ContractFactory;
 import org.aion.rlp.RLP;
 import org.aion.rlp.RLPElement;
 import org.aion.rlp.RLPItem;
@@ -24,8 +28,8 @@ import org.aion.types.Address;
 import org.aion.types.ByteArrayWrapper;
 
 public class AionContractDetailsImpl extends AbstractContractDetails {
-
     private ByteArrayKeyValueStore dataSource;
+    private ByteArrayKeyValueStore objectGraphSource = null;
 
     private byte[] rlpEncoded;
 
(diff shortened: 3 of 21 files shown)


modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -1342,6 +1342,21 @@ private AionBlockSummary applyBlock(IAionBlock block) {
...
+                AionTransaction tx = receipt.getTransaction();
+                if (tx.isContractCreationTransaction() && receipt.isSuccessful()) {
+                    track.saveVmType(
+                            tx.getContractAddress(),
+                            TransactionTypeRule.isValidAVMContractDeployment(tx.getTargetVM())
+                                    ? TransactionTypes.AVM_CREATE_CODE
+                                    // FVM contracts do not always have the correct type
+                                    : TransactionTypes.FVM_CREATE_CODE);
+                    if (tx.getTargetVM() != 1) {
+                        System.out.println("Contract creation transaction receipt: " + receipt);
```
AionJayT: use log instead of the std output?

modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
```diff
@@ -110,18 +118,80 @@ public ByteArrayWrapper get(ByteArrayWrapper key) {
...
+
+        this.objectGraph = graph;
+        this.objectGraphHash = h256(objectGraph);
+
+        this.setDirty(true);
+        this.rlpEncoded = null;
+    }
+
+    // TODO: use until the AVM impl is added
+    private final boolean enableObjectGraph = false;
```
AionJayT: eventually, the AVM will use the ObjectGraph. I think the flag can be removed.

modVM/src/org/aion/vm/KernelInterfaceForAVM.java
```diff
@@ -76,26 +80,32 @@ public void putCode(Address address, byte[] code) {
...
-        return new byte[0];
+        return this.repositoryCache.getObjectGraph(contract);
     }
 
     @Override
     public void putStorage(Address address, byte[] key, byte[] value) {
         ByteArrayWrapper storageKey = new ByteArrayWrapper(key);
         ByteArrayWrapper storageValue = new ByteArrayWrapper(value);
         this.repositoryCache.addStorageRow(address, storageKey, storageValue);
+        if (this.repositoryCache.getVmType(address) != AVM_CREATE_CODE) {
```
AionJayT: the check looks weird (also the L106) What's the reason for forcing the storage change the vm type?

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
AlexandraRoatis: sorry, this commit was not meant to be added to the PR; I removed it

modVM/src/org/aion/vm/KernelInterfaceForAVM.java
AlexandraRoatis: Each kernel interface is called only if the contract fit its definition. Whenever the call fits a contract behavior like adding storage, code or the object graph I check if the VM code was not set yet. Otherwise, we cannot ensure that the correct code is set until the blockchain checks for contract transactions. Without this setting, we cannot guarantee that a contract can be correctly deployed and called in the same block.

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -376,6 +376,18 @@ public ByteArrayWrapper getStorageValue(Address address, ByteArrayWrapper key) {
         return (details == null) ? EMPTY_BYTE_ARRAY : details.getCode(codeHash);
     }
 
+    @Override
+    public byte getVmType(Address contract) {
```
aionick: Is it possible to return `TransactionTypes` instead here? Just so that the caller gets a more meaningful object in their hands?

modVM/src/org/aion/vm/BulkExecutor.java
```diff
@@ -449,4 +450,22 @@ private boolean isContractAddress(Address address) {
...
+
+    private byte getVmType(Address destination) {
+        if (ContractFactory.isPrecompiledContract(destination)) {
+            // skip the call to disk
+            return TransactionTypes.FVM_CREATE_CODE;
+        } else {
+            byte storedVmType = repositoryChild.getVMUsed(destination);
+
+            // DEFAULT is returned when there was no contract information stored
+            if (storedVmType == TransactionTypes.DEFAULT) {
```
aionick: Again, is this supposed to be an error?

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -1342,6 +1342,18 @@ private AionBlockSummary applyBlock(IAionBlock block) {
...
                 receipts.add(summary.getReceipt());
                 summaries.add(summary);
+
+                // save contract creation data to contract details
+                AionTxReceipt receipt = summary.getReceipt();
+                AionTransaction tx = receipt.getTransaction();
+                if (tx.isContractCreationTransaction() && receipt.isSuccessful()) {
+                    track.saveVmType(
+                            tx.getContractAddress(),
+                            TransactionTypeRule.isValidAVMContractDeployment(tx.getTargetVM())
```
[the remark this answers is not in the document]
AlexandraRoatis: Unfortunately, there are contracts deployed on the `mastery` network with the incorrect VM type, so we need the TransactionTypeRule to validate AVM/FVM contracts differently depending on the fork parameter.

modAionImpl/src/org/aion/zero/impl/db/AbstractContractDetails.java
```diff
@@ -23,8 +23,11 @@
     protected int detailsInMemoryStorageLimit;
 
     private Map<ByteArrayWrapper, byte[]> codes = new HashMap<>();
-    // set to FVM_CREATE_CODE to update encoding for FVM contracts
-    protected byte vmType = TransactionTypes.FVM_CREATE_CODE;
+    // classes extending this rely on this value starting off as null
+    protected byte[] objectGraph = null;
+
+    // using the default transaction type to specify undefined VM
+    protected byte vmType = TransactionTypes.DEFAULT;
```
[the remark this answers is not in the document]
AlexandraRoatis: This is not the only purpose. This value will also be used by transactions that are not contract deployments.

modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
```diff
@@ -110,18 +118,80 @@ public ByteArrayWrapper get(ByteArrayWrapper key) {
...
+    @Override
+    public byte[] getObjectGraph() {
+        if (objectGraph == null) {
+            // the object graph was not stored yet
+            if (java.util.Arrays.equals(objectGraphHash, EMPTY_DATA_HASH)) {
+                return EMPTY_BYTE_ARRAY;
+            } else {
+                // note: the enforced use of optional is rather cumbersome here
+                Optional<byte[]> dbVal = getContractObjectGraphSource().get(objectGraphHash);
+                objectGraph = dbVal.isPresent() ? dbVal.get() : null;
```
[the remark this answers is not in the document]
AlexandraRoatis: `null` means undefined, while empty byte array can be a valid setting; when the value is `null` I may need to check for the parent's object graph and load it

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
AlexandraRoatis: I will refactor the `TrsansactionTypes` into a enum since they are currently constants.

modVM/src/org/aion/vm/KernelInterfaceForAVM.java
AlexandraRoatis: I don't like it either, but so far there is no such method. We need to add a `createContract` to the interfaces that would have to be called whenever a contract is created. This way we can also avoid instantiating and managing contract details objects for all the created accounts that most often are not actual contracts.

modVM/src/org/aion/vm/KernelInterfaceForAVM.java
```diff
@@ -193,6 +205,19 @@ public boolean destinationAddressIsSafeForThisVM(Address address) {
...
         // Otherwise, it must be an Avm contract address.
-        return TransactionTypeRule.isValidAVMContractDeployment(repositoryCache.getVMUsed(address));
+        return TransactionTypeRule.isValidAVMContractDeployment(getVmType(address));
+    }
+
+    private byte getVmType(Address destination) {
+        byte storedVmType = repositoryCache.getVMUsed(destination);
+
+        // DEFAULT is returned when there was no contract information stored
+        if (storedVmType == TransactionTypes.DEFAULT) {
```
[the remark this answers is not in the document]
AlexandraRoatis: The problem is that contract details objects get created in the case of simple accounts that are not contracts as well. This requires a lot of refactoring to clean up the implementation. The refactoring is delegated to 0.4.1 release though.

modVM/src/org/aion/vm/BulkExecutor.java
AlexandraRoatis: This value will be returned for contracts that have not been stored, resulting from databases that have been migrated to the new build without re-importing all blocks. In that case, we check for the vm type stored in the contract details object.

PR #867, opened 2019-04-04
FVM contract encoding update

AlexandraRoatis:
## Description

1. Bug fix: closing the database where the VM type is stored and storing the information only if the contract creation was successful.
2. Added VM type to contract details and saved it to stored encoding assuming only FVM contracts exist on the network.

Subset of #866.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

- existing test suite + tests to verify encoding switch

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
@@ -147,10 +147,10 @@ public class AionContractDetailsImpl extends AbstractContractDetails {
 
         RLPList rlpList = (RLPList) data.get(0);
 
-        if (rlpList.size() == 5) {
-            // compatible with old encoding
-            decodeEncodingWithoutVmType(rlpList, fastCheck);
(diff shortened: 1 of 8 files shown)


modAionImpl/src/org/aion/zero/impl/db/AionContractDetailsImpl.java
```diff
@@ -181,13 +216,71 @@ public void decode(byte[] rlpCode, boolean fastCheck) {
             externalStorage = true;
             storageTrie.getCache().setDB(getExternalStorageDataSource());
         }
+    }
+
+    /**
+     * Decodes the new version of encoding which is a list of 6 elements, specifically:<br>
+     * { 0:address, 1:vmType, 2:isExternalStorage, 3:storageRoot, 4:storage, 5:code }
```
AionJayT: the rlp encode/decode order can be the original order plus the vmType. Then we can make sure the logic will be more neat and safe.

PR #879, opened 2019-04-15
Post-fork transaction validation

AlexandraRoatis:
## Description

- Set the `AVM_CREATE_CODE = 0x02`;
- Allowing only transactions with types 0, 1, 2 after the fork;
- Unit tests for transaction validation before and after the fork.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

- existing tests
- two new tests to verify correct transaction validation after the fork

## Verification

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
@@ -21,9 +21,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.aion.interfaces.block.Constant;
-import org.aion.interfaces.db.Repository;
-import org.aion.interfaces.db.RepositoryCache;
 import org.aion.evtmgr.IEvent;
 import org.aion.evtmgr.IEventMgr;
 import org.aion.evtmgr.IHandler;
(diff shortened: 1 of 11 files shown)


modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
```diff
@@ -441,7 +441,7 @@ public synchronized TxResponse addPendingTransaction(AionTransaction tx) {
     }
 
     public boolean isValid(AionTransaction tx) {
```
AionJayT: this method can be a static method?

modMcf/src/org/aion/mcf/types/AbstractTransaction.java
```diff
@@ -54,8 +54,12 @@ public AbstractTransaction(byte[] nonce, Address receiveAddress, byte[] value, b
         this.to = receiveAddress;
         this.value = value;
         this.data = data;
-        // default type 0x01; reserve date for multi-type transaction
-        this.type = TransactionTypes.DEFAULT;
+        // setting the default type depending on use case
+        if (to == null) {
+            this.type = TransactionTypes.FVM_CREATE_CODE;
```
AionJayT: How does the AVM contract create definition?
I thought the AVM contract create address destination is `null` too

AlexandraRoatis: For AVM contracts the default value is incorrect. However, our unit tests have already been written to specify the deploy code and not use the default value when testing the AVM contract deployments. This change is mainly for the sake of the old FVM tests that were written without taking the transaction type into account.

AlexandraRoatis: I would prefer fully removing the constructor that allows building transactions without specifying the type, but that would be a low priority refactoring that can be applied later.

modAionImpl/src/org/aion/zero/impl/blockchain/AionPendingStateImpl.java
AlexandraRoatis: It cannot be changed to static because of the interface for the pending state.

AionJayT: okay.

modMcf/src/org/aion/mcf/types/AbstractTransaction.java
AionJayT: thanks for the explaination

PR #881, opened 2019-04-16
fix p2p taskStatus disappear issue

AionJayT:
## Notice

## Description

- This PR is trying to solve the thread p2p-ts stuck issue. Currently, the p2p-ts is using scheduled threads pool with the other thread task. I suspect the p2p-ts stuck due to the `reqnodestasks` thread blocking the pool. Therefore, I use dedicated thread instead of using the sharing thread pool to avoid the blocking.

Fixes Issue # .

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

- Modified the existing test case with new implementation

## Verification

### DIFF
--- a/modP2pImpl/src/org/aion/p2p/impl/TaskRequestActiveNodes.java
+++ b/modP2pImpl/src/org/aion/p2p/impl/TaskRequestActiveNodes.java
@@ -23,6 +23,7 @@ public final class TaskRequestActiveNodes implements Runnable {
     public void run() {
         INode node = mgr.getRandom();
         if (node != null) {
+            Thread.currentThread().setName("p2p-reqNodes");
             if (p2pLOG.isTraceEnabled()) {
                 p2pLOG.trace("TaskRequestActiveNodes: {}", node.toString());
             }
--- a/modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
+++ b/modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
@@ -226,8 +226,9 @@ public final class P2pMgr implements IP2pMgr {
             }
 
             if (p2pLOG.isInfoEnabled()) {
-                scheduledWorkers.scheduleWithFixedDelay(
-                        getStatusInstance(), 2, PERIOD_SHOW_STATUS, TimeUnit.MILLISECONDS);
+                Thread threadStatus = new Thread(getStatusInstance(), "p2p-ts");
+                threadStatus.setPriority(Thread.NORM_PRIORITY);
+                threadStatus.start();
             }
 
             if (!syncSeedsOnly) {
@@ -469,7 +470,8 @@ public final class P2pMgr implements IP2pMgr {
     }
 
     private TaskStatus getStatusInstance() {
-        return new TaskStatus(this.nodeMgr, this.selfShortId, this.sendMsgQue, this.receiveMsgQue);
+        return new TaskStatus(
+                this.start, this.nodeMgr, this.selfShortId, this.sendMsgQue, this.receiveMsgQue);
     }
 
     private TaskClear getClearInstance() {
--- a/modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
+++ b/modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
@@ -462,20 +462,19 @@ public final class P2pMgr implements IP2pMgr {
     }
 
     private TaskSend getSendInstance(int i) {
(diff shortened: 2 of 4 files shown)


modP2pImpl/src/org/aion/p2p/impl/TaskRequestActiveNodes.java
```diff
@@ -23,6 +23,7 @@ public TaskRequestActiveNodes(final IP2pMgr _mgr, final Logger p2pLOG) {
     public void run() {
         INode node = mgr.getRandom();
         if (node != null) {
+            Thread.currentThread().setName("p2p-reqNodes");
```
aion-kelvin: since we have to set the thread's name, would it make sense for TaskRequestActiveNodes to extend Thread instead of implement Runnable?  I feel like we're assuming things about which thread this Runnable will be in, so why not just make it a Thread?

alternatively, make the thread-name-setting happen in P2pMgr line 231 instead (set the name before starting threadStatus)

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskStatus.java
```diff
@@ -21,21 +26,28 @@ public TaskStatus(
...
+                    p2pLOG.debug(
+                        "recv queue[{}] send queue[{}]",
+                        this.receiveMsgQue.size(),
+                        this.sendMsgQue.size());
+                } else if (p2pLOG.isInfoEnabled()) {
+                    p2pLOG.info(status);
+                }
+            } catch (Exception e) {
+                p2pLOG.warn("P2p taskStatus exception! ", e);
+            }
```
AlexandraRoatis: please also add a message when exiting the loop to know that the thread shut down

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskStatus.java
```diff
@@ -21,21 +26,28 @@ public TaskStatus(
...
+                    p2pLOG.debug(status);
+                    p2pLOG.debug(
+                        "recv queue[{}] send queue[{}]",
+                        this.receiveMsgQue.size(),
+                        this.sendMsgQue.size());
+                } else if (p2pLOG.isInfoEnabled()) {
+                    p2pLOG.info(status);
+                }
+            } catch (Exception e) {
+                p2pLOG.warn("P2p taskStatus exception! ", e);
```
AlexandraRoatis: here we should print a message only if `start.get()` is not true since the task could have been interrupted for shutdown

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
```diff
@@ -469,7 +470,8 @@ private TaskReceive getReceiveInstance() {
     }
 
     private TaskStatus getStatusInstance() {
-        return new TaskStatus(this.nodeMgr, this.selfShortId, this.sendMsgQue, this.receiveMsgQue);
+        return new TaskStatus(
+                this.start, this.nodeMgr, this.selfShortId, this.sendMsgQue, this.receiveMsgQue);
```
AlexandraRoatis: I think we could omit all the `this.` and keep the code in one line

modP2pImpl/src/org/aion/p2p/impl/TaskRequestActiveNodes.java
AionJayT: I think the purpose to use runnable is for saving the memory overhead, cause the task only executing `run()` method. you don't need to create a lot of objects by extending it to a Thread.

AionJayT: Set the thead name mainly for more readable when you are using jstack to debug the thread info.

modP2pImpl/src/org/aion/p2p/impl1/P2pMgr.java
AionJayT: updated

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskStatus.java
AionJayT: added!

modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskStatus.java
AionJayT: added  InterruptedException handle

PR #924, opened 2019-07-04
Remove vm_api dependency

aionick:
## Notice

It is not allowed to submit your PR to the 'master' branch directly, please submit your PR to the 'master-pre-merge' branch and rebase your PR to 'master-pre-merge' branch.

## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-


modTxPool/src/org/aion/txpool/Constant.java
```diff
@@ -0,0 +1,6 @@
+package org.aion.txpool;
+
+public class Constant {
+
+    public static final int MAX_BLK_SIZE = 2 * 1024 * 1024;
+}
```
AionJayT: The Constant class is part of a consensus for limiting the block size. Should not be part of txpool. Can move to the other place in the following modularity stuff. (This number can be set in the txpool during the init stage.)

aionick: I will make a note of this on Jira. This was the only option for now, since `modAionImpl` and `modTxPoolImpl` both use this class but neither of them are dependent on each other, they're both dependent on `modTxPool` however.

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
```diff
@@ -58,13 +58,13 @@
 import org.aion.rlp.RLP;
 import org.aion.types.AionAddress;
 import org.aion.util.bytes.ByteUtil;
+import org.aion.util.bytes.Hash256;
```
AionJayT: For Hash256,
I would prefer it to stay in the org.aion.util.types package.

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryCache.java
```diff
@@ -12,18 +12,18 @@
...
 import org.aion.mcf.core.AccountState;
+import org.aion.mcf.db.ContractDetails;
 import org.aion.mcf.db.IBlockStoreBase;
+import org.aion.mcf.db.InternalVmType;
+import org.aion.mcf.db.Repository;
+import org.aion.mcf.db.RepositoryCache;
 import org.aion.precompiled.ContractFactory;
 import org.aion.types.AionAddress;
-import org.aion.vm.api.types.ByteArrayWrapper;
+import org.aion.util.types.ByteArrayWrapper;
```
AlexandraRoatis: This is also a good candidate for `aion_types`, but this can be left as a later TODO.

modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
aionick: Done.

Issue #560, opened 2018-07-24
Feature request: re-import all blocks


iamyulong:
# Introduction
`reindex`, to rebuild the database by re-importing all existing blocks, is a useful feature. It allows node operators to quickly clean up the database without doing a sync with the network; it also enables developers to verify if their changes will break any consensus rule.

# Current situation
This feature is missing in the current Aion implementation.

# Task Description
We're asking for contributors to implement this feature for https://github.com/aionnetwork/aion. This task can be divided into the following steps:
- Extract all blocks from the current database folders;
- Wipe the whole database (or move to another temporary place);
- Rebuild the index and state by importing all the extracted blocks;
- Verify if the whole process is successful or not;
- Expose this feature by extending the currently command line interface.

Note: we're open to alternative approaches.

# Acceptance Criteria
- The code implements the targets described above
- The code is well documented and fully covered by unit tests
- The code is compliant with Aion's code convention
- The code gets merged into the main code repository

# Required Skill
Participants needs to have experience in Java software development, and have basic knowledge about blockchain system.

# Guidance
Technical support is available through [Gitter](https://gitter.im/aionnetwork/Lobby).

qoire:
@kzeine, this is a good candidate for bounty

AlexandraRoatis:
Most of this functionality is already implemented between the `./aion.sh -r` and `./aion.sh --state FULL/SPREAD/TOP`.

AlexandraRoatis:
PR #800 introduces a new CLI command `--redo-import` / `--redo-import <height>`.
This command triggers a process of reimporting all the blocks from the given height (or from the genesis block when no parameter is provided). The` state`, `details`, `storage` and `transaction` databases are completely wiped and rebuilt when starting from the genesis block. The `index` database is expected to have the correct main chain blocks information for this task to complete.

Future enhancements:
- [ ] fully rebuild the `index` database as well;
- [ ] improve time estimates by approximating and factoring in the number of keys added to the state;
- [ ] improve execution speed by batching IO operations;
- [ ] improve code by using `nanoTime` and `TimeUnit`.

PR #929, opened 2019-07-11
Re #560: --redo-import now rewrites the index and prunes side chains

AlexandraRoatis:
## Description

- Corrected the bug where multiple index entries were introduced for the same block during the rebuild.
- Also rewriting the `index` database and removing side chain blocks from the `block` database.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

- Ran the command on a correct database with side chains and one with incorrect index entries. Ensured that both databases are correct after the `--redo-import` execution with start from the genesis block and start from a block > 0.

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
@@ -17,6 +17,7 @@ import java.math.BigInteger;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -1250,6 +1251,36 @@ public void correctMainChain(AionBlock block, Logger log) {
...
+
+                // delete all the side-chain blocks
+                for (BlockInfo blockInfo : level) {
+                    if (!Arrays.equals(currentHash, blockInfo.getHash())) {
+                        blocks.delete(blockInfo.getHash());
+                    }
+                }
+
+                // replace all the block info with empty list
+                index.set(block.getNumber(), Collections.emptyList());
```
AionJayT: Looks like it only reset the index db at the reimport point level, where is the logic to reset the level between the reimport point to the current index DB height?

AlexandraRoatis: before the block import, the index is cleared and side-chains are pruned, then the block is imported which sets the correct index value

AlexandraRoatis: the index is corrected for each block that is imported, it does not modify the database for blocks predating the first block to re-import

AionJayT: Okay. understood.

PR #960, opened 2019-08-20
IExternalState block difficulty is now BigInteger

aionick:
This is the AVM's external state interface.


aionick:
For some reason the commits are listed in the wrong order in this PR view. Better to look at it directly on the branch. The difference is that `c7bc053` comes before `6d901ad` (this is the final commit), that's all.

PR #961, opened 2019-08-20
Optimizations and caching

AlexandraRoatis:
## Description

- Optimized check for stored blocks by using the cached index size;
- Bugfix and optimized state pruning:  
    - using a cache to retrieve the block numbers and hashes for pruning;
    - updated method signatures to take only needed data and avoid unnecessary byte array conversions;
    - bugfix: all blocks at each level will be pruned;
- Correctly closing the store instead of directly closing the database;
- ObjectDataSource alternatives with caching and debugging features: 
    - cached versions using LRU and Caffeine implementations;
    - new builder class for creating different instances;
    - new `CACHE` logging level for getting cache statistics;
- AKI-263: Cache imported blocks to increase performance;
- AKI-275: Cache contract index data + skip database access on empty code;

Fixes Issue AKI-263 and AKI-275.

## Type of change

- [x] Bug fix.
- [x] New feature.
- [x] Enhancement.
- [ ] Unit test.

### DIFF
--- a/modMcf/build.gradle
+++ b/modMcf/build.gradle
@@ -26,6 +26,7 @@ dependencies {
     compile group: 'org.apache.commons', name: 'commons-collections4', version: '4.0'
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.4'
     compile 'org.json:json:20180813'
+    compile 'com.github.ben-manes.caffeine:caffeine:2.8.0'
 
     testCompile project(':modAionImpl')
     testCompile 'junit:junit:4.12'
--- a/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/AionBlockchainImpl.java
@@ -1199,7 +1199,7 @@ public class AionBlockchainImpl implements IAionBlockchain {
             }
             transactionStore.flushBatch();
 
-            repository.commitBlock(block.getHeader());
+            repository.commitBlock(block.getHashWrapper(), block.getNumber(), block.getStateRoot());
 
             if (LOG.isDebugEnabled())
                 LOG.debug(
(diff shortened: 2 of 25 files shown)


modMcf/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
     compile group: 'org.apache.commons', name: 'commons-collections4', version: '4.0'
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.4'
     compile 'org.json:json:20180813'
+    compile 'com.github.ben-manes.caffeine:caffeine:2.6.2'
```
ben-manes: can you use 2.8.0?

In 2.7 we made the policy adaptive so that it perform better in LRU-biased workloads.

ben-manes:
Please let me know if you have a workload where Caffeine underperforms or LRU is superior. With the improvements in 2.7, this should not occur (be within a margin of error). I need to do some minor tuning for tiny caches (e.g. 50 and under) but small (200+) were excellent.

AlexandraRoatis:
@ben-manes Thank you for the suggestion! I will definitely update the library. The two cases I updated in this PR are predictably LRU-biased and require only tiny caches of 10 items to get around 98% utilization. However, I am working on other pending changes where a small cache would be the preferred option.
Also, the stats provided by the Caffeine cache were very useful in my experiments. Thank you very much for that!

ben-manes:
great. I didn't get around to tuning the hill climber algorithm for super tiny caches, so it is LFU-biased in those cases. It's an easy fix but, with no users complaining, it has been in my backlog. Corda is a blockchain trace that is LRU-biased and Caffeine nails it now, but again the maximum has to be a few 200+ for adaptivity to kick in. Otherwise the step size is too small and it converges too soon, leaving the cache in an LFU configuration.

PR #966, opened 2019-08-22
AKI-326: Remove external module dependencies from p2p

AlexandraRoatis:
## Description

- Removed the dependency on aion.log and aion.util from aion.p2p.impl:
    - The `Logger` object is now passed as input to constructors and methods.
    - The `Node` class implements its own hex string conversion with unit tests to ensure correctness.

Fixes Issue AKI-326.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.


AlexandraRoatis:
I added 3 more commits, one to detach modApiServer from p2p and the other two to clean up some of the exposed p2p functionality.

PR #974, opened 2019-09-03
Added getBlockdDetails method [AKI-351]

dev-dennis-guye:
## Description

This change adds a new ops_getBlockDetails method which returns a blockdetails similar to the blockdetails found in the Java API.

Fixes Issue # AKI-351

## Type of change
- [ ] Bug fix.
- [x] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing
- deployed the build and synced to mainnet
- executed requests using blocks that contained no transaction, one transaction and more than one transaction. Requests made: 
```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetails","params": [1]}'
```

```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetails","params": {"block":6753}}'
```

```
curl 192.168.1.67:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetails","params": [1174759]}'
```
- added functional test cases to the dashboard ETL

### DIFF
--- a/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
@@ -13,6 +13,7 @@ import java.math.BigInteger;
 import java.math.RoundingMode;
 import java.time.Instant;
 import java.time.ZoneOffset;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
@@ -2267,6 +2268,57 @@ public class ApiWeb3Aion extends ApiAion {
         return new RpcMsg(result);
     }
 
+    public RpcMsg ops_getBlockDetailsByNumber(Object _params){
+        String _blockNumber;
+        if (_params instanceof JSONArray) {
+            _blockNumber = ((JSONArray) _params).get(0) + "";
+        } else if (_params instanceof JSONObject) {
+            _blockNumber = ((JSONObject) _params).get("block") + "";
+        } else {
+            return new RpcMsg(null, RpcError.INVALID_PARAMS, "Invalid parameters");
+        }
+        AionBlock block;
+        Long bn = this.parseBnOrId(_blockNumber);
+
+        // user passed a Long block number
+        if (bn != null && bn >= 0) {
+            block = (AionBlock) this.ac.getBlockchain().getBlockByNumber(bn);
+            if (block == null) {
+                return new RpcMsg(JSONObject.NULL);
+            }
+        }else {
+            return new RpcMsg(JSONObject.NULL);
+        }
+
+        return rpcBlockDetailsFromBlock(block);
+    }
+
+
+    public RpcMsg ops_getBlockDetailsByHash(Object _params){
+        String _blockHash;
+        if (_params instanceof JSONArray) {
+            _blockHash = ((JSONArray) _params).get(0) + "";
+        } else if (_params instanceof JSONObject) {
+            _blockHash = ((JSONObject) _params).get("block") + "";
+        } else {
+            return new RpcMsg(null, RpcError.INVALID_PARAMS, "Invalid parameters");
+        }
+
+        Block block = this.ac.getBlockchain().getBlockByHash(ByteUtil.hexStringToBytes(_blockHash));
+
+        if (block == null) {
+                return new RpcMsg(JSONObject.NULL);
+        }
+
+        Block mainBlock = this.ac.getBlockchain().getBlockByNumber(block.getNumber());
+
+        if (mainBlock == null || !Arrays.equals(block.getHash(), mainBlock.getHash())) {
+            return new RpcMsg(JSONObject.NULL);
+        }
+
+        return rpcBlockDetailsFromBlock(block);
+    }
+
     public RpcMsg ops_getBlock(Object _params) {
         String _bnOrHash;
         boolean _fullTx;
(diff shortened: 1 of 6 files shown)


modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2267,6 +2268,70 @@ public RpcMsg ops_getTransaction(Object _params) {
...
+                if (block == null) {
+                    return new RpcMsg(JSONObject.NULL);
+                }
+            } else {
+                return new RpcMsg(JSONObject.NULL);
+            }
+        }
+
+        // see if the user passed in a hash
+        if (block == null) {
```
aion-kelvin: I'm not a big fan of trying to interpret the argument as a block number and then trying to interpret it as a hash if the block number was not found.  Can we have two methods -- getBlockDetailsByHash and getBlockDetailsByNumber?

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2267,6 +2268,70 @@ public RpcMsg ops_getTransaction(Object _params) {
...
+            block = (AionBlock) this.ac.getBlockchain().getBlockByHash(ByteUtil.hexStringToBytes(_bnOrHash));
+            if (block == null) {
+                return new RpcMsg(JSONObject.NULL);
+            }
+        }
+
+        Block mainBlock = this.ac.getBlockchain().getBlockByNumber(block.getNumber());
```
aion-kelvin: Is there a reason we shouldn't return block details for non-main-chain blocks?

modApiServer/src/org/aion/api/server/types/Blk.java
```diff
@@ -94,6 +96,36 @@ public static Object AionBlockToJson(
...
+
+        JSONObject obj = AionBlockOnlyToJson(genericBlock, totalDifficulty);
+
+        if(obj == null){
+            return null;
+        }
+
+        long blockTime;
+        if (previousTimestamp == null){
+            blockTime = 0;
```
aion-kelvin: would null make more sense than 0 here?

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2267,6 +2268,70 @@ public RpcMsg ops_getTransaction(Object _params) {
...
+            AionTxInfo txInfo = this.ac.getAionHub()
+                .getBlockchain()
+                .getTransactionInfo(transaction.getTransactionHash());
+            txInfoList.add(txInfo);
+        }
+
+
+        return new RpcMsg(
+            Blk.aionBlockDetailsToJson(block,
+                txInfoList, previousBlock==null ? null:previousBlock.getTimestamp(), totalDiff, blkReward));
```
aion-kelvin: what causes previousBlock == null?

dev-dennis-guye: This case occurs if the requested block is the genesis block.

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
dev-dennis-guye: There is no reason why we shouldn't but since the change is specifically designed for the dashboard I added the check since the dashboard only displays mainchain blocks.

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
dev-dennis-guye: Ok that makes sense I will update the implementation.

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
aion-kelvin: Sounds reasonable

modApiServer/src/org/aion/api/server/types/Blk.java
dev-dennis-guye: I updated the implementation to return 0 if the requested block was the genesis block and otherwise return null

aion-kelvin:
Looks good.  One last thing -- can you also check that the following all give the same result?  Looking at the code, I think they do, but just want to see it work in real life ... if these are all equivalent, then let's merge it:

`curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"jsonrpc":"2.0","method":"ops_getBlockDetails","params": [11]}'`

`curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"jsonrpc":"2.0","method":"ops_getBlockDetails","params": ["11"]}'`

`curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"jsonrpc":"2.0","method":"ops_getBlockDetails","params": ["0xb"]}'`

dev-dennis-guye:
The change was re-tested using the following requests and re-running the tests in the dashboard ETL code.

```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetailsByNumber","params": {"block":"0x1"}}' | jq  -M
```

```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetailsByNumber","params": {"block": 1}}' | jq  -M
```

```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetailsByNumber","params": [1]}' | jq  -M
```

```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetailsByNumber","params": ["0x1"]}' | jq  -M
```

```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetailsByHash","params": [0x28375cc29c3ba1b76f36de5a486eac814b566f488c6c98a30dfea854f7f2671a]}' | jq  -M
```

```
curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"ops_getBlockDetailsByHash","params": {"block":0x28375cc29c3ba1b76f36de5a486eac814b566f488c6c98a30dfea854f7f2671a}}' | jq  -M
```

PR #994, opened 2019-09-26
removed bad casts in RPC methods[AKI-409]

dev-dennis-guye:
## Description

Removes casts from the generic Block type to the POW AionBlock type which caused failures in several RPC methods

Fixes Issue # AKI-409

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing
Ran from the command line `curl localhost:8545 -H 'Content-Type: application/json' -s -d '{"json-Mc":"2.0","method":"eth_getBlockByNumber","params": {"block":82707}}' | jq
`

### DIFF
--- a/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
@@ -50,6 +50,7 @@ import org.aion.evtmgr.IEventMgr;
 import org.aion.evtmgr.IHandler;
 import org.aion.evtmgr.impl.callback.EventCallback;
 import org.aion.evtmgr.impl.evt.EventTx;
+import org.aion.mcf.blockchain.BlockHeader.BlockSealType;
 import org.aion.zero.impl.keystore.Keystore;
 import org.aion.mcf.blockchain.Block;
 import org.aion.mcf.config.CfgApi;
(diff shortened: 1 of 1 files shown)


modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2648,8 +2660,7 @@ public RpcMsg stratum_getdifficulty() {
     }
 
     public RpcMsg stratum_getmininginfo() {
-        // TODO: [Unity] This cast should be removed when we support staking blocks
-        AionBlock bestBlock = (AionBlock) getBestBlock();
+        Block bestBlock =  getBestBlock();
```
AionJayT: I suspect this API should leave for the mining block, not staking block. Therefore, it probably requires to find the previous mining block instead of bestblock

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2744,15 +2755,25 @@ public RpcMsg stratum_getHeaderByBlockNumber(Object _params) {
...
+                if (block != null){
+                    if (block.getHeader().getSealType().equals(BlockSealType.SEAL_POW_BLOCK)) {
+                        A0BlockHeader header = (A0BlockHeader) block.getHeader();
+                        obj.put("code", 0); // 0 = success
+                        obj.put("nonce", toHexString(header.getNonce()));
+                        obj.put("solution", toHexString(header.getSolution()));
+                        obj.put("headerHash", toHexString(header.getMineHash()));
+                        obj.putOpt("blockHeader", header.toJSON());
+                    } else if(block.getHeader().getSealType().equals(BlockSealType.SEAL_POS_BLOCK)) {
+                        StakingBlockHeader header = ((StakingBlock)block).getHeader();
```
AionJayT: should remove the staking block case

PR #1000, opened 2019-10-02
Sync refactoring to improve performance

AlexandraRoatis:
## Description

Rewrote the sync algorithm to use a header request manager that keeps track of peer states and received headers. Simplifies sync state management and improves performance.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.


modP2pImpl/src/org/aion/p2p/impl1/tasks/TaskSend.java
```diff
@@ -62,89 +66,205 @@ public void run() {
...
+            case OUTBOUND:
+                node = nodeMgr.getOutboundNode(mo.getNodeId());
+                break;
+        }
+
+        if (node != null) {
+            // delay sending the request if we have multiple peers and the last sent message was to
+            // the current peer by placing the request in the pendingRequests that will be processed
+            // during the next iteration
+            if (mgr.getActiveNodes().size() > 1 && node.getIdHash() == lastNode.getIdHash()) {
```
AionJayT: not a big concern, the size check can just use  " ! isEmpty()"

AlexandraRoatis: Actually, here the difference is between 1 or larger. If only 1 peer exists, we will not enforce the requirement of not sending another request to the same peer.

AionJayT: ah, okay.

PR #1027, opened 2019-10-22
Transformed code fix

geoff-aion:
## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-


modMcf/src/org/aion/mcf/db/TransformedCodeInfo.java
```diff
@@ -0,0 +1,43 @@
...
+
+        ByteArrayWrapper transformedCode = avmVersionMap.get(avmVersion);
+
+        return transformedCode == null ? null : transformedCode.toBytes();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
AlexandraRoatis: We could make this class final and replace the above line with `if (o instanceof TransformedCodeInfo) {`

modMcf/src/org/aion/mcf/db/TransformedCodeInfo.java
```diff
@@ -0,0 +1,43 @@
...
+
+        if (avmVersionMap == null) return null;
+
+        ByteArrayWrapper transformedCode = avmVersionMap.get(avmVersion);
+
+        return transformedCode == null ? null : transformedCode.toBytes();
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
AlexandraRoatis: When implementing `equals` we also need `hashCode`. Where is this used? If possible I would not implement it at all and rely on object equality.

modMcf/src/org/aion/mcf/db/TransformedCodeInfo.java
```diff
@@ -0,0 +1,43 @@
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        TransformedCodeInfo transformedCodeInfo = (TransformedCodeInfo) o;
+        return this.transformedCodeMap.equals(transformedCodeInfo.transformedCodeMap);
```
AlexandraRoatis: I don't think this performs equality as intended.

modMcf/src/org/aion/mcf/db/TransformedCodeInfo.java
```diff
@@ -0,0 +1,43 @@
+package org.aion.mcf.db;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.aion.util.types.ByteArrayWrapper;
+
+public class TransformedCodeInfo {
+
+    // Key for this map is the codeHash, the value is a map which has AVM version as its key and transformed code as its value.
+    public Map<ByteArrayWrapper, Map<Integer, ByteArrayWrapper>> transformedCodeMap;
```
AlexandraRoatis: The values for the inner map should be `byte[]` since they are not used as keys, the wrapper would only decrease performance.

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -213,37 +219,24 @@ public void updateBatch(
                 }
             }
 
-            if (LOG.isTraceEnabled()) {
-                LOG.trace("updated: detailsCache.size: {}", detailsCache.size());
+            for (Map.Entry<AionAddress, TransformedCodeInfo> entry : transformedCodeCache.entrySet()) {
+                transformedCodeSource.put(entry.getKey().toByteArray(), entry.getValue());
```
AlexandraRoatis: I think this should be calling `setTransformedCode` otherwise it might overwrite old entries

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -413,15 +406,45 @@ public ByteArrayWrapper getStorageValue(AionAddress address, ByteArrayWrapper ke
...
+                .get(address.toByteArray());
+
+            if (transformedCodeInfo == null) {
+                transformedCodeInfo = new TransformedCodeInfo();
+            }
+
+            transformedCodeInfo.add(ByteArrayWrapper.wrap(codeHash), avmVersion,
+                ByteArrayWrapper.wrap(transformedCode));
+            transformedCodeSource.put(address.toByteArray(), transformedCodeInfo);
+            transformedCodeSource.commit();
```
AlexandraRoatis: this commit should be moved into the repository's flush method

modAionImpl/src/org/aion/zero/impl/db/AionRepositoryCache.java
```diff
@@ -41,10 +42,15 @@
 
     protected ReadWriteLock lockDetails = new ReentrantReadWriteLock();
 
+    /** local transformed code cache */
+    protected Map<AionAddress, TransformedCodeInfo> cachedTransformedCode;
+    protected ReadWriteLock lockTransformedCode = new ReentrantReadWriteLock();
```
AlexandraRoatis: Please remove this lock and use the full write lock method (for both read and write). Introducing another lock is too risky here, e.g. `updateBatch` is missing the necessary lock. I will open an issue to reduce locking to a single object in this class. The current use of multiple locks is risky due to the write required by loading the account state and contract details and not particularly useful since the cache is rarely accessed by multiple threads.

PR #1050, opened 2019-10-29
Avm is initialized in a single place

aionick:
- We had a problem where the avm was being initialized in multiple places
  for different CLI arguments, and was even not being initialized at times
  it should be. This change fixes these issues by initializing the avm in
  a single place that all CLI variations have to pass through.
- This change also fixes AKI-457 by checking that the thread owns the avm
  lock before calling release on it first.

## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-

### DIFF
--- a/modBoot/src/org/aion/Aion.java
+++ b/modBoot/src/org/aion/Aion.java
@@ -408,7 +408,9 @@ public class Aion {
         } catch (IOException e) {
             // Nothing to handle here, we are shutting down...
         } finally{
-            AvmProvider.releaseLock();
+            if (AvmProvider.holdsLock()) {
+                AvmProvider.releaseLock();
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


modBoot/src/org/aion/Aion.java
```diff
@@ -82,6 +82,7 @@ public static void main(String args[]) {
 
         ReturnType ret = new Cli().call(args, cfg);
         if (ret != ReturnType.RUN) {
+            shutdownAvm();
```
AlexandraRoatis: We can call `System.exit` from other parts of the code. Perhaps a better place for the AVM shutdown would be in the `ShutdownHook` thread defined below (around line 350).

aionick: Ideally we register the hook first thing and it just knows whether a component has been started/loaded and how to shut it down if so, but as it stands that's one of the last things we run in the main method because we wait for everything to be started up first

AlexandraRoatis: Ah, I see. Looks good to me.

PR #1074, opened 2019-11-18
Filtered out POS blocks from the minerStatsView data store[AKI-513].

dev-dennis-guye:
## Description
Filters POS blocks from the POW block cache.

Fixes Issue # AKI-513.

## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing
- A curl request from the terminal

### DIFF
--- a/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
@@ -22,6 +22,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReadWriteLock;
@@ -2722,7 +2723,7 @@ public class ApiWeb3Aion extends ApiAion {
     private class MinerStatsView {
 
         LinkedList<byte[]> hashQueue; // more precisely a dequeue
-        Map<byte[], Block> blocks;
+        Map<byte[], AionBlock> blocks;
         private JSONObject response;
         private int qSize;
         private byte[] miner;
@@ -2756,7 +2757,7 @@ public class ApiWeb3Aion extends ApiAion {
 
             int blkTimesAccumulated = 0;
             Long lastBlkTimestamp = null;
-            Block b = null;
+            AionBlock b = null;
 
             try {
                 // index 0 = latest block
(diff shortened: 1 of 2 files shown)


modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2817,7 +2818,7 @@ private JSONObject buildResponse() {
 
         MinerStatsView update() {
             // get the latest head
-            Block blk = getBestBlock();
+            Block blk = getBestMiningBlock();
```
AionJayT: We should force the call return AionBlock because it always returns a mining block

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2828,8 +2829,8 @@ MinerStatsView update() {
             }
 
             // evict data as necessary
-            LinkedList<Map.Entry<byte[], Block>> tempStack = new LinkedList<>();
-            tempStack.push(Map.entry(blk.getHash(), blk));
+            LinkedList<Map.Entry<byte[], AionBlock>> tempStack = new LinkedList<>();
+            tempStack.push(Map.entry(blk.getHash(), (AionBlock) blk));
```
AionJayT: So you can remove the casting here

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -2849,8 +2850,11 @@ MinerStatsView update() {
                     && blk.getNumber() > 2) {
 
                 blk = getBlockByHash(blk.getParentHash());
-                tempStack.push(Map.entry(blk.getHash(), blk));
-                itr++;
+                if (blk.getHeader().getSealType().equals(BlockSealType.SEAL_POW_BLOCK)) {
```
AionJayT: Can remove the check also.

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
dev-dennis-guye: Agreed but the blk variable must use the base type because the variable may also hold staking blocks.

AionJayT: If you use the getBestMiningBlock call, the kernel should guarantee you always get the AionBlcok.So you need to modify the return of the ApiAion class.

dev-dennis-guye: The blk variable is used as a temporary variable, like this: `blk = getBlockByHash(blk.getParentHash());`, which could return a pos block.

PR #1078, opened 2019-12-02
Rpc generation

dev-dennis-guye:
## Description Added additional methods to generated RPC library

- Includes all methods in personal and any methods in ops used by the Dashboard.

- expansion of ChainHolder interface
- refactored Account Manage to handle all account interactions

Fixes Issue # AKI-533 and AKI-534.

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.

## Testing

- unit tests in OpsRPCImplTest and PersonalRPCImplTest


modApiServer/src/org/aion/api/server/rpc3/Web3EntryPoint.java
```diff
@@ -50,6 +51,11 @@ public String call(String requestString){
         RpcError err = null;
         Integer id = null;
         Object resultUnion = null;
+        Stopwatch stopwatch = null;
+        if (logger.isDebugEnabled()){
```
AionJayT: If the logger got pass from the AionLogger, then doesn't need this check anymore.

modApiServer/src/org/aion/api/server/rpc3/Web3EntryPoint.java
```diff
@@ -66,17 +72,24 @@ public String call(String requestString){
...
+            logger.error("Request failed due to an internal error: ", e);
+            err= new InternalErrorRPCException(
+                e.getClass().getSimpleName()+":"+e.getMessage()).getError(); // Prefer this over
+                                                                // creating an instance of RpcError
         }
         final String resultString = ResponseConverter
             .encodeStr(new Response(id, resultUnion, err, VersionType.Version2));
-        logger.debug("Produced response: {}", resultString);
+
+        if (logger.isDebugEnabled() && stopwatch != null) {
```
AionJayT: If the logger got pass from the AionLogger, then doesn't need this check anymore.

modApiServer/test/org/aion/api/server/rpc3/MiningRPCImplTest.java
```diff
@@ -12,6 +12,8 @@
 
 import java.math.BigInteger;
 import java.util.function.Function;
+import org.aion.api.server.account.AccountManager;
```
AionJayT: Is it an unnecessary import? because we have the interface.

AionJayT:
For the AionChainHolder class, is it any better to create a blockchain interface instead of involving the core implementation classes in it? I suspect we are trying to decouple the modules dependency as we can.

modApiServer/src/org/aion/api/server/rpc3/Web3EntryPoint.java
dev-dennis-guye: We are using the check to avoid creating an instance of the stopwatch if it is not necessary.

modApiServer/src/org/aion/api/server/rpc3/Web3EntryPoint.java
dev-dennis-guye: I suppose here the check could be `stopwatch != null`

dev-dennis-guye:
> For the AionChainHolder class, is it any better to create a blockchain interface instead of involving the core implementation classes in it? I suspect we are trying to decouple the modules dependency as we can.


The idea is to eventually move the implementation of this class and the account manager outside of the rpc module. We can then inject the implementation when building the RPC server.

arajasek:
Looks good to me, for the most part. I had one question about commit  80ec64f.
Also, do we have a better sense of which methods in ops are actually needed? I think this could be a good juncture to eliminate unused methods (instead of introducing them, and later scrapping them).

modApiServer/src/org/aion/api/server/rpc3/Web3EntryPoint.java
AionJayT: I mean inside the AionLogger, it already checked the isDebugEnabled() call. So you don't have to check in the code if this is an AionLogger.

dev-dennis-guye:
> I feel like I don't understand the purpose of this commit. How does creating this interface help with testing? I'm somewhat opposed to adding unnecessary interfaces.

Final classes cannot be mocked by mockito, so rather than removing the final modifier on the account manager class I created the interface that can be used for mocking.

dev-dennis-guye:
> Looks good to me, for the most part. I had one question about commit [80ec64f](https://github.com/aionnetwork/aion/commit/80ec64f8536d84931cb58d692d11d21e5ddf446c).
> Also, do we have a better sense of which methods in ops are actually needed? I think this could be a good juncture to eliminate unused methods (instead of introducing them, and later scrapping them).

Based on my experience with the ETL we are only using ops_getBlockDetailsByNumber and ops_getAccountState. Most of the other methods were not included.

AionJayT:
> For the AionChainHolder class, is it any better way to create a blockchain interface instead of involving the core implementation classes in it? I suspect we are trying to decouple the modules dependency as we can.



> > I feel like I don't understand the purpose of this commit. How does creating this interface help with testing? I'm somewhat opposed to adding unnecessary interfaces.
> 
> Final classes cannot be mocked by mockito, so rather than removing the final modifier on the account manager class I created the interface that can be used for mocking.



> > For the AionChainHolder class, is it any better to create a blockchain interface instead of involving the core implementation classes in it? I suspect we are trying to decouple the modules dependency as we can.
> 
> The idea is to eventually move the implementation of this class and the account manager outside of the rpc module. We can then inject the implementation when building the RPC server. However, I do not think we are at that point yet.

It's not just the accountMgr, but also the blockchain core component including repository.
Yap, we should use injection to decouple the module.

arajasek:
All right, sounds good to me :)

PR #1079, opened 2019-12-02
Added new links

Sheikh-Aman:
Left Whitepapers link unchanged because we don't have them hosted anywhere yet

## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.

## Testing

-

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,6 @@ To keep up to date and stay connected with current progress and development, rea
 [Aion on Twitter](https://twitter.com/Aion_OAN)  
 [Aion Blog](https://blog.aion.network/)
 
-For more information about Aion Community please refer to [Aion Community](https://aion.network/community/)
 
 ## Join the Team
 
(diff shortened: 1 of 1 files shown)


AlexandraRoatis:
I think the link labels need to be revised as well, for example:
- "Mining Docs" -> "Validators Docs"
- "Aion" -> "OAN" where appropriate

Sheikh-Aman:
> I think the link labels need to be revised as well, for example:
> 
> * "Mining Docs" -> "Validators Docs"
> * "Aion" -> "OAN" where appropriate

Good catch!
I just updated the mining references to include validation too, and at some places keeping them both, to not confuse the people who have known mining

Aion ->OAN won't be applicable in these cases as far as I can see, since this works closely with validator experience, instead of developer experience. But correct me please if I am wrong.

PR #1080, opened 2019-12-02
AKI-601 Limit the eth_getLogs can't query more than 1000 blocks

AionJayT:
## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-

### DIFF
--- a/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
+++ b/modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
@@ -1,6 +1,7 @@
 package org.aion.api.server.rpc;
 
 import static java.util.stream.Collectors.toList;
+import static org.aion.api.server.types.FltrLg.BLOCKS_QUERY_MAX;
 import static org.aion.util.bytes.ByteUtil.EMPTY_BYTE_ARRAY;
 import static org.aion.util.conversions.Hex.toHexString;
 import static org.aion.util.types.HexConvert.hexStringToBytes;
@@ -25,8 +26,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Function;
 import org.aion.api.server.ApiAion;
 import org.aion.api.server.ApiTxResponse;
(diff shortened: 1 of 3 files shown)


modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -1164,12 +1163,20 @@ private FltrLg createFilter(ArgFltr rf) {
...
         if (bnTo != BEST_PENDING_BLOCK && bnFrom > bnTo) {
             LOG.debug("jsonrpc - eth_newFilter(): from block is after to block");
-            return null;
+            filter.setFilterError("jsonrpc - eth_newFilter(): from block is after to block");
+            return filter;
+        }
+
+        if (bnTo >= (bnFrom + BLOCKS_QUERY_MAX)) {
+            LOG.debug("jsonrpc - eth_newFilter(): can't query more than 1000 blocks");
+            filter.setFilterError("jsonrpc - eth_newFilter(): can't query more than 1000 blocks");
```
AlexandraRoatis: Both messages above should use the constant value instead of `1000` so that the message stays consistent even when the limit changes.

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -1220,8 +1227,8 @@ public RpcMsg eth_newFilter(Object _params) {
         }
 
         FltrLg filter = createFilter(rf);
-        if (filter == null) {
-            return new RpcMsg(null, RpcError.INVALID_PARAMS, "Invalid filter object provided.");
+        if (filter.getFilterError() != null) {
+            return new RpcMsg(null, RpcError.INVALID_PARAMS, "Invalid filter object provided." + filter.getFilterError());
```
AlexandraRoatis: Minor: a space is needed between the two concatenated messages; same applies to the `if` below

modApiServer/src/org/aion/api/server/rpc/ApiWeb3Aion.java
```diff
@@ -1164,12 +1163,20 @@ private FltrLg createFilter(ArgFltr rf) {
 
         if (bnFrom == null || bnTo == null) {
             LOG.debug("jsonrpc - eth_newFilter(): from, to block parse failed");
-            return null;
+            filter.setFilterError("jsonrpc - eth_newFilter(): from, to block parse failed");
+            return filter;
```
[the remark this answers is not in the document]
AionJayT: added the comment before the if condition.

PR #1091, opened 2019-12-06
AKI-516: Quickly return null when asked to create impossible block te…

arajasek:
…mplates

- This prevents users from seeing many messages like "Tried to create 2 PoS blocks in a row!"
- It also reduces unnecessary work

## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-


arajasek:
Closing this until we fix an issue with the mining pool interaction, and then confirm that this commit works as intended.

PR #1101, opened 2019-12-17
AKI-603(647) AionPendingState refactoring

AionJayT:
## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
@@ -302,15 +302,25 @@ public class AionBlockchainImpl implements IAionBlockchain {
             this.forkUtility.enableUnityFork(maybeUnityFork.get());
         }
 
+        Optional<Long> maybe040Fork = load040ForkNumberFromConfig(CfgAion.inst());
+        if (maybe040Fork.isPresent()) {
+            if (maybe040Fork.get() < 0) {
+                LOG.warn("The 040 fork number cannot be less than 0, set the fork number to 0");
+                maybe040Fork = Optional.of(0L);
+            }
+
+            this.forkUtility.enable040Fork(maybe040Fork.get());
+        }
+
         // initialize beacon hash validator
         this.beaconHashValidator = new BeaconHashValidator(this, this.forkUtility);
     }
 
     /**
-     * Determine fork 0.5.0 fork number from Aion Config.
+     * Determine fork 1.0 fork number from Aion Config.
      *
      * @param cfgAion configuration
-     * @return 0.5.0 fork number, if configured; {@link Optional#empty()} otherwise.
+     * @return 1.0 fork number, if configured; {@link Optional#empty()} otherwise.
      * @throws NumberFormatException if "fork1.0" present in the config, but not parseable
      */
     private static Optional<Long> loadUnityForkNumberFromConfig(CfgAion cfgAion) {
@@ -319,6 +329,22 @@ public class AionBlockchainImpl implements IAionBlockchain {
             return Optional.empty();
         } else {
             return Optional.of(Long.valueOf(unityforkSetting));
+        }
+    }
+
+    /**
+     * Determine fork 0.4.0 fork number from Aion Config.
+     *
+     * @param cfgAion configuration
+     * @return 0.4.0 fork number, if configured; {@link Optional#empty()} otherwise.
+     * @throws NumberFormatException if "fork0.4.0" present in the config, but not parseable
+     */
+    private static Optional<Long> load040ForkNumberFromConfig(CfgAion cfgAion) {
+        String fork040Setting = cfgAion.getFork().getProperties().getProperty("fork0.4.0");
+        if(fork040Setting == null) {
+            return Optional.empty();
+        } else {
+            return Optional.of(Long.valueOf(fork040Setting));
         }
     }
 
--- a/modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
@@ -8,6 +8,7 @@ import static org.aion.util.biginteger.BIUtil.isMoreThan;
 import static org.aion.util.conversions.Hex.toHexString;
 
 import java.util.EnumMap;
+import org.aion.zero.impl.blockchain.AionHub.BestBlockImportCallback;
 import org.aion.zero.impl.blockchain.AionHub.SelfNodeStatusCallback;
 import org.aion.zero.impl.core.IDifficultyCalculator;
 import static org.aion.zero.impl.core.ImportResult.EXIST;
@@ -197,6 +198,7 @@ public class AionBlockchainImpl implements IAionBlockchain {
         .synchronizedMap(new LRUMap<>(64));
 
     private SelfNodeStatusCallback callback;
+    private BestBlockImportCallback bestBlockCallback;
 
     public AionBlockchainImpl(CfgAion cfgAion, boolean forTest) {
         this(generateBCConfig(cfgAion), AionRepositoryImpl.inst(),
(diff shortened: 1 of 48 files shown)


modAionImpl/src/org/aion/zero/impl/pendingState/AionPendingStateImpl.java
```diff
@@ -1253,4 +1203,21 @@ public void updateBest() {
...
+        this.networkBestBlockCallback = callback;
+    }
+
+    private boolean isCloseToNetworkBest() {
+        // This case is trying to avoid the NPR during the unitTest (because no AionImpl instance been init)
+        if (networkBestBlockCallback == null) {
+            return true;
+        }
+
+        return best.get().getNumber() >= (networkBestBlockCallback.getNetworkBestBlockNumber() - 128);
```
AionJayT: make 128 as a variable

modAionImpl/src/org/aion/zero/impl/blockchain/AionImpl.java
```diff
@@ -67,6 +67,7 @@ private AionImpl(boolean forTest) {
 
         blockchainCallbackInterfaces = Collections.synchronizedList(new ArrayList<>());
         aionHub.getPendingState().setPendingTxCallback(new PendingTxCallback(blockchainCallbackInterfaces));
+        aionHub.getPendingState().setNetworkBestBlockNumberCallback(new NetworkBestBlockCallback(this));
```
AlexandraRoatis: I think this should be moved to the pending state constructor. This would remove the need for the setter and the NPE worries.

modAionImpl/src/org/aion/zero/impl/blockchain/AionImpl.java
```diff
@@ -68,6 +67,7 @@ private AionImpl(boolean forTest) {
         blockchainCallbackInterfaces = Collections.synchronizedList(new ArrayList<>());
         aionHub.getPendingState().setPendingTxCallback(new PendingTxCallback(blockchainCallbackInterfaces));
         aionHub.getPendingState().setNetworkBestBlockNumberCallback(new NetworkBestBlockCallback(this));
+        aionHub.getPendingState().setTxBroadcastCallback(new TxBroadcastCallback(this));
```
AlexandraRoatis: Similar to previous comments, I think this should be included in the pending state constructor.

modAionImpl/src/org/aion/zero/impl/blockchain/AionImpl.java
```diff
@@ -379,6 +370,21 @@ public void pendingTxStateUpdateCallback(PendingTxDetails txDetails) {
...
+        IAionChain chainInterface;
+
+        TxBroadcastCallback(IAionChain chainInterface) {
+            if (chainInterface == null) {
+                throw new NullPointerException();
+            }
+            this.chainInterface = chainInterface;
+        }
+
+        public void broadcastTx(List<AionTransaction> transactions) {
```
AlexandraRoatis: minor: I'd prefer the name `broadcastTransactions`

modAionImpl/src/org/aion/zero/impl/blockchain/AionHub.java
```diff
@@ -117,7 +117,7 @@ private void initializeHub(
         this.blockchain = _blockchain == null ? new AionBlockchainImpl(cfg, forTest) : _blockchain;
         blockchain.setEventManager(this.eventMgr);
 
-        this.mempool = AionPendingStateImpl.create(cfg, blockchain, forTest);
+        this.mempool = forTest ? AionPendingStateImpl.createForTest(cfg, blockchain) : AionPendingStateImpl.create(cfg, blockchain);
```
AlexandraRoatis: I don't like this change. We should be increasing the functionality that can be tested not separating and reducing it further.

modAionImpl/src/org/aion/zero/impl/blockchain/AionHub.java
```diff
@@ -133,7 +133,7 @@ private void initializeHub(
         }
 
         this.startingBlock = this.blockchain.getBestBlock();
-        if (!cfg.getConsensus().isSeed()) {
+        if (!cfg.getTx().isSeedMode()) {
```
AlexandraRoatis: This change will be problematic when migrating to the release. I am not sure if it's worth the trouble that it may cause. Can we introduce some sort of config migration to avoid the need for manual user update?

modAionImpl/src/org/aion/zero/impl/pendingState/AionPendingStateImpl.java
```diff
@@ -119,57 +118,79 @@ private synchronized void backupPendingTx() {
...
+                        poolDump);
+        ps.testingMode = true;
+        ps.poolDumpEnable = !seedMode;
         return ps;
     }
 
-    private AionPendingStateImpl(CfgAion _cfgAion) {
-
-        this.isSeed = _cfgAion.getTx().isSeedMode();
+    private AionPendingStateImpl(
```
AlexandraRoatis: Could we make this constructor public and avoid the need for the two static factory methods?

modAionImpl/test/org/aion/zero/impl/sync/BlockPropagationTest.java
```diff
@@ -341,7 +341,14 @@ public void send(int _nodeId, String s, Msg _msg) {
                         anotherBundle.bc.getBlockHeaderValidator(),
                         false,
                         (byte) 2,
-                        AionPendingStateImpl.createForTest(CfgAion.inst(), anotherBundle.bc));
+                        AionPendingStateImpl.createForTest(
+                                anotherBundle.bc,
+                                CfgAion.inst().getConsensus().getEnergyStrategy().getUpperBound(),
```
AlexandraRoatis: I don't think we should use `CfgAion.inst()` to set this up. The test should have it's own values for these settings.

modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
```diff
@@ -1717,15 +1717,15 @@ private boolean isValid(Block block) {
...
+                                    TXValidator.validateTx(tx, unityForkEnabled).isFail()
                                                 || !TransactionTypeValidator.isValid(tx)
                                                 || !beaconHashValidator.validateTxForBlock(tx, block))) {
                     LOG.error("Some transactions in the block are invalid");
 
                     for (AionTransaction tx : txs) {
                         TX_LOG.debug(
                                 "Tx valid ["
-                                        + TXValidator.isValid(tx, unityForkEnabled)
+                                        + !TXValidator.validateTx(tx, unityForkEnabled).isFail()
```
AlexandraRoatis: we should have a `isSuccess()` in the `TxResponse`

modAionImpl/src/org/aion/zero/impl/blockchain/AionHub.java
AionJayT: No, because the seed mode is a hidden config option. So I don't think it will have an issue for migrating.

AionJayT: I can keep the hidden config option in the cfgConsesus (and add comments), then update the logic to CfgTx

modAionImpl/src/org/aion/zero/impl/sync/handler/BroadcastTxHandler.java
```diff
@@ -81,38 +73,23 @@ public final void receive(int _nodeIdHashcode, String _displayId, final byte[] _
...
-            return;
-        }
 
-        try {
             for (AionTransaction tx : castRawTx(broadCastTx)) {
-                if (!txQueue.offer(tx)) {
-                    if (log.isTraceEnabled()) {
-                        log.trace("<BroadcastTxHandler txQueue full! {}>", _displayId);
-                    }
+                if (!txQueue.offer(tx, 100, TimeUnit.MILLISECONDS)) {
```
AionJayT: remove buffer time. give it a log

PR #1107, opened 2020-01-21
AKI-646 switch default DB to the rocksDB

AionJayT:
## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/DatabaseUtils.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/DatabaseUtils.java
@@ -8,12 +8,15 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Objects;
 import java.util.Properties;
+import java.util.stream.Stream;
 import org.aion.db.impl.ByteArrayKeyValueDatabase;
 import org.aion.db.impl.DatabaseFactory;
 import org.aion.db.impl.DatabaseFactory.Props;
 import org.aion.db.impl.PersistenceMethod;
 import org.aion.mcf.db.exception.InvalidFilePathException;
+import org.aion.mcf.db.exception.InvalidFileTypeException;
 import org.slf4j.Logger;
 
 /** @author Alexandra Roatis */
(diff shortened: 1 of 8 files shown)


modAionImpl/src/org/aion/zero/impl/db/DatabaseUtils.java
```diff
@@ -8,12 +8,17 @@
 import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Objects;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
```
AlexandraRoatis: I think some of these imports are not needed.

networks/amity/config.xml
```diff
@@ -66,9 +66,9 @@
         <state-storage>FULL</state-storage>
         <!--Database implementation used to store data; supported options: leveldb, h2, rocksdb.-->
         <!--Caution: changing implementation requires re-syncing from genesis!-->
-        <vendor>leveldb</vendor>
+        <vendor>rocksdb</vendor>
         <!--Boolean value. Enable/disable database compression to trade storage space for execution time.-->
-        <enable_db_compression>false</enable_db_compression>
+        <enable_db_compression>true</enable_db_compression>
```
AlexandraRoatis: For LevelDB the compression made performance fairly bad. Do we have experimental data to show that it's better to enable compression with RocksDB?

AionJayT: I can do the benchmark. But bast on my understanding, Enable the data compression will increase CPU usage but significantly reduce the data size been wrote into the disk, especially the compacting happens. Also I doubt the leveldb library might use the old version of the native levelDB, So the compression performance is not good. for the benchmark I did before. You can look at the AKI-550. 
Also,you can look at the reference link for the details of the rocksdb data compression
https://github.com/facebook/rocksdb/wiki/Compression

AionJayT: Also, the default setting in rust kernel also enable the data compression.
https://github.com/aionnetwork/aionr/blob/master/db/core/src/dbconfigs.rs#L82-L100

PR #1109, opened 2020-01-23
AKI-648 validate block timestamp in the block submit api

AionJayT:
## Description

-

Fixes Issue # .

## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

## Testing

-


modApiServer/src/org/aion/api/server/rpc3/AionChainHolder.java
```diff
@@ -261,4 +269,8 @@ private Repository getRepoByBlockNumber(long blockNumber){
         return this.chain.getRepository()
             .getSnapshotTo(getBlockByNumber(blockNumber).getStateRoot());
     }
+
+    private static boolean isFutureTimestamp(long timestamp) {
+        return timestamp > (System.currentTimeMillis() / 1000 + 1);
```
AlexandraRoatis: Could we use the `FutureBlockRule` here? It would make the code more maintainable and devs less likely to update one implementation and not the other.

AionJayT:
> Are all the updated jars in the `tooling` directory necessary?
No, but the new tooling has less issue for creating staking block in the custom network. Also new tooling only has one jar instead of the externalStaker use a lot of  jars

modApiServer/src/org/aion/api/server/rpc3/AionChainHolder.java
AionJayT: updated to use the `FutureBlockRule`

PR #1122, opened 2020-03-04
Blockstore use clean-up

AlexandraRoatis:
## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.


modAionImpl/src/org/aion/zero/impl/blockchain/AionHub.java
```diff
@@ -125,7 +120,7 @@ private void initializeHub(
         blockchain.setEventManager(this.eventMgr);
 
         try {
-            loadBlockchain();
+            startingBlock = this.blockchain.load(cfg.getGenesis(), genLOG);
```
AionJayT: the line 131 can be removed.

PR #1126, opened 2020-03-05
AKI-672: Call repository close on main instance

AlexandraRoatis:
## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [ ] Unit test.

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
+++ b/modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
@@ -133,6 +133,7 @@ import org.slf4j.LoggerFactory;
 public class AionBlockchainImpl implements IAionBlockchain {
 
     private static final Logger LOG = LoggerFactory.getLogger(LogEnum.CONS.name());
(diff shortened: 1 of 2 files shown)


modAionImpl/src/org/aion/zero/impl/blockchain/AionBlockchainImpl.java
```diff
@@ -2186,6 +2187,14 @@ public synchronized void setBestBlock(Block block) {
 
     @Override
     public synchronized void close() {
+        while (repository.isSnapshot()) {
+            popState();
```
AionJayT: Does the kernel flush the DB before the shutdown process? If the answer is no, why we do popState?

AlexandraRoatis: The main repository instance is stashed when the snapshot is created. If the current repository is a snapshot that means the main one is in the stack. So we need to pop the stack until we get to the main repository instance that contains access too all the databases that must be closed.
If the close was called under normal circumstances the repository was already flushed. If it was called due to an error (like a VM issue) then flushing may store corrupt data, so it shouldn't be done.

PR #1129, opened 2020-03-11
Unit tests for the Blockstore

AlexandraRoatis:
## Type of change

- [x] Bug fix.
- [ ] New feature.
- [ ] Enhancement.
- [x] Unit test.

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
+++ b/modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
@@ -1429,18 +1429,19 @@ public class AionBlockStore {
     }
 
     /**
-     * Retrieve two generation blocks with unity protocol info with one lock.
+     * Retrieves two generation blocks with unity protocol info.
+     * <p>
+     * Always returns a 2-element array. If the blocks cannot be retrieved the array will contain null values.
+     * Block[0] is the parent block and has the given hash. Block[1] is the grandparent block.
      *
-     * @param hash given hash of the block
-     * @return the 2 generation block data have matched hash with unity protocol info. Block[0] is
-     *     the parent block, Block[1] is the grandParent block. The return might only contain the
-     *     parent block and still return the 2-elements array.
+     * @param hash the hash of the parent block
+     * @return the retrieved two generation blocks with unity protocol info
      */
     public final Block[] getTwoGenerationBlocksByHashWithInfo(byte[] hash) {
-        if (hash == null) {
-            return null;
-        }
         Block[] blockFamily = new Block[] { null, null};
+        if (hash == null) {
+            return blockFamily;
+        }
 
         lock.lock();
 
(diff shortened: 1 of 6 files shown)


modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -1392,18 +1392,21 @@ void redoIndexWithoutSideChains(Block block) {
...
+     * Block[2] is the great grandparent block.
+     *
+     * @param hash the hash of the parent block
+     * @return the retrieved three generation blocks with unity protocol info
      */
     public final Block[] getThreeGenerationBlocksByHashWithInfo(byte[] hash) {
+        Block[] blockFamily = new Block[] { null, null, null};
         if (hash == null) {
-            return null;
+            return blockFamily;
```
AionJayT: I think this change will let AionBlockChainImpl.isValid() throw null exception. Please have a check.

modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
```diff
@@ -1429,18 +1429,19 @@ void redoIndexWithoutSideChains(Block block) {
...
-     *     the parent block, Block[1] is the grandParent block. The return might only contain the
-     *     parent block and still return the 2-elements array.
+     * @param hash the hash of the parent block
+     * @return the retrieved two generation blocks with unity protocol info
      */
     public final Block[] getTwoGenerationBlocksByHashWithInfo(byte[] hash) {
+        Block[] blockFamily = new Block[] { null, null};
         if (hash == null) {
-            return null;
+            return blockFamily;
```
AionJayT: Will let AionBlockChainImpl.createNewMiningBlockInternal() throw NPE

AlexandraRoatis: It would already throw and an NPE because the blockFamily is accessed without a null check.

modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
AlexandraRoatis: I updated the `null` check in `AionBlockchainImpl.isValid` to exit if the parent block is null.

AionJayT: Cool.

modAionImpl/src/org/aion/zero/impl/db/AionBlockStore.java
AionJayT: So We should just throw NPE if the input hash equal to null. instead of return values (or return null)

Same as the other method.

AlexandraRoatis: Ok. I made the update.

PR #1132, opened 2020-03-17
Pruning Tests and Bug Fixes

AlexandraRoatis:
## Type of change

- [x] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [x] Unit test.

### DIFF
--- a/modAionImpl/src/org/aion/zero/impl/blockchain/AionHub.java
+++ b/modAionImpl/src/org/aion/zero/impl/blockchain/AionHub.java
@@ -393,14 +393,12 @@ public class AionHub {
     // is ahead of the oldBlockTemplate
     // Returns null if we're waiting on a Staking block, or if creating a new block template failed for some reason
     public BlockContext getNewMiningBlockTemplate(BlockContext oldBlockTemplate, long systemTime) {
-        if (blockchain.isUnityForkEnabledAtNextBlock() &&
-                blockchain.getBestBlock().getHeader().getSealType() == BlockHeader.BlockSealType.SEAL_POW_BLOCK) {
-            return null;
-        } else {
-            BlockContext context;
-            
-            blockTemplateLock.lock();
-            try {
+        blockTemplateLock.lock();
+        try {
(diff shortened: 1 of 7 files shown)


modAionImpl/src/org/aion/zero/impl/db/AionRepositoryImpl.java
```diff
@@ -512,10 +512,6 @@ public void flush() {
                         db.commit();
                     }
                 }
-            } else {
```
AionJayT: What scenario we will see the databaseGroup equal to null ? Sometimes We can see this message during the seednode running, also the project test case.

AlexandraRoatis: The databaseGroup is null when using a repository snapshot. The snapshot is always used when importing a side chain block. The message is really useless, cause it doesn't notify us of anything meaningful.

modAionImpl/src/org/aion/zero/impl/blockchain/AionHub.java
```diff
@@ -427,14 +427,14 @@ public BlockContext getNewMiningBlockTemplate(BlockContext oldBlockTemplate, lon
...
+        Block best = blockchain.getBestBlock();
+        if (best.getHeader().getSealType() == BlockHeader.BlockSealType.SEAL_POS_BLOCK) {
             return null;
         } else {
             StakingBlock blockTemplate;
             blockTemplateLock.lock();
             try {
-                blockTemplate = blockchain.createStakingBlockTemplate(
-                        mempool.getPendingTransactions(), signingPublicKey, newSeed, coinbase);
+                blockTemplate = blockchain.createStakingBlockTemplate(best, mempool.getPendingTransactions(), signingPublicKey, newSeed, coinbase);
```
AionJayT: Should we just use blockTemplateLock and the try...finally to include all of the method?
Same as the getMiningBlockTemplate

AlexandraRoatis: I added an extra commit for this change.

PR #1141, opened 2020-04-13
AKI-692: Refactor Sync manager to use ExecutorService 

AlexandraRoatis:
## Type of change

- [ ] Bug fix.
- [ ] New feature.
- [x] Enhancement.
- [ ] Unit test.


modAionImpl/src/org/aion/zero/impl/sync/SyncMgr.java
```diff
@@ -380,6 +394,25 @@ private void interruptAndWait(Thread t, long timeout) {
...
 
+    private void shutdownAndAwaitTermination(ExecutorService pool) {
+        pool.shutdown(); // Disable new tasks from being submitted
+        try {
+            // Wait a while for existing tasks to terminate
+            if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
+                pool.shutdownNow(); // Cancel currently executing tasks
+                // Wait a while for tasks to respond to being cancelled
+                if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
+                    log.error("Pool did not terminate");
```
AionJayT, line 459: Should we throw the exception in this if condition?

modAionImpl/src/org/aion/zero/impl/sync/SyncMgr.java
```diff
@@ -393,11 +439,32 @@ public long getNetworkBestBlockNumber() {
     }
 
     public synchronized void shutdown() {
+        if (p2pLog.isDebugEnabled()) {
```
AionJayT, line 425: minor: the check can be removed

modAionImpl/src/org/aion/zero/impl/sync/SyncMgr.java
AlexandraRoatis: Would an exception provide any additional information? I can add a stack trace to the log if you think it's useful?

modAionImpl/src/org/aion/zero/impl/sync/SyncMgr.java
AlexandraRoatis: The check here is actually an optimization to skip the other if statements when debug is disabled. Would you still want it removed?

modAionImpl/src/org/aion/zero/impl/sync/SyncMgr.java
AionJayT: To throw the exception is not necessary, it mainly try to interrupt the thread. If the kernel thread stuck, we can dump the process threads info.

modAionImpl/src/org/aion/zero/impl/sync/SyncMgr.java
AionJayT: I see, because you call the method in the log argument. Just ignore my comment.

